PR #3, opened 2018-08-29
Use --schema-only when calling pg_dump

stephenpaulger:
`pg_dump` will dump the entire contents of a database table without the `--schema-only` option. In `DumpTableDDL()` `COPY` statements are ignored, so there should be no side effect to not generating them in the first place.

The `COPY` command that `pg_dump` generates to dump the table's content doesn't work on Redshift databases, so this PR is also a step towards letting sqldef work on Redshift databases.


stephenpaulger:
Thanks.

PR #13, opened 2018-09-21
Add AFTER and FIRST

kawakami-o3:
I have added AFTER and FIRST to 'ALTER TABLE ... ADD COLUMN'.
It can make the sequence of columns same with its CREATE statement.


k0kubun:
The feature looks good!

The DDL looks like mysql-specific. Could you switch the behavior by `g.mode`?

kawakami-o3:
Done. Please check my modifications.

PR #16, opened 2018-10-15
Create Dockerfile

exussum12:
build with 

docker build -t sqldef .

then run with 

 docker run -i sqdef to run mysql version

docker run -i sqdef ./psqldef
to run psql version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM golang:1.11.1-alpine3.8
+run apk add --no-cache build-base make gcc  git; \
+  git clone https://github.com/k0kubun/sqldef.git; \
+  export GOPATH=/go/ ; \
+  export GOBIN=$HOME/bin ; \
+  cd sqldef; \
+  make all;
+
+WORKDIR "/go/sqldef/build/linux-amd64"
+CMD ["/go/sqldef/build/linux-amd64/mysqldef"]


k0kubun:
I don't have any reason to reject this, but I don't think of any motivation to support this either. Could you describe why committing your Dockerfile to this repository is beneficial for you? 

Why not maintain Dockerfile on your own repository or somewhere else which you can easily handle? At least I would use sqldef in binary release form and so the Dockerfile wouldn't be actively maintained by me even if it were committed to this upstream.

k0kubun:
Thank you for your interest to this project anyways!

exussum12:
The use case is stateless build machines. This means either wget'ing the binary (knowing the arch) or just building. Im trying to use this tool as part of a database build / maintenance script, ie checked in database structures and using git to apply the correct files.  

The Dockerfile makes the building simple and consistent across architectures.

The file was something I was using and thought that other developers would find it useful also. Ideally it would be pushed to dockerhub but it does work without that also.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.11.1-alpine3.8
+run apk add --no-cache build-base make gcc  git; \
```
k0kubun: Run this separately from `COPY` commented above.

k0kubun:
> The file was something I was using and thought that other developers would find it useful also.

Your opinion makes sense, but actually I feel your Dockerfile is designed for your own internal use and not a suitable thing to be committed to this repository.

Anyway, I'll merge your work and fix commented parts so that your goal "other developers would find it useful" is met.

exussum12:
Looks great. The git requirement can also be removed now. (apk line)

k0kubun:
I tried the same thing but fetching go dependencies from GitHub seems to require that.

PR #23, opened 2018-11-20
MySQL timestamp initialization "ON UPDATE CURRENT_TIMESTAMP"

tsuty:
MySQL "ON UPDATE CURRENT_TIMESTAMP"  support.

"ON UPDATE" is already parsed. 
https://github.com/k0kubun/sqldef/blob/6ae81394adbb0b2d854237e7a9e2db0f08835fac/sqlparser/parser.y#L720

I know that "defaultVal, length, scale" vales are difficult to compare.
But  "ON UPDATE" value is easy to compare.


k0kubun:
Thanks! :+1:

PR #25, opened 2019-06-10
supporting enum values

### LINKED ISSUE #24: Support Enum values
At least on mysqldef, appending the column with enum values fails.

When you would like to add the definition of enum column like 
`source_type enum('a','b') NOT NULL DEFAULT 'a' AFTER yyy;`,
the expected `ALTER` statement is 
`ALTER TABLE xxx ADD COLUMN source_type enum('a','b') NOT NULL DEFAULT 'a' AFTER yyy;`.

But for now, sqldef generates the statement like 
`ALTER TABLE xxx ADD COLUMN source_type enum NOT NULL DEFAULT 'a' AFTER yyy;`

The lack of the definition of enum values causes a runtime error.

timakin:
Fixes #24

### DIFF
--- a/cmd/mysqldef/mysqldef_test.go
+++ b/cmd/mysqldef/mysqldef_test.go
@@ -517,6 +517,29 @@ func TestMysqldefOnUpdate(t *testing.T) {
 	assertApplyOutput(t, createTable, nothingModified)
 }
 
+func TestMysqldefEnumValues(t *testing.T) {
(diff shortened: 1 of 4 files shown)


k0kubun:
Thanks :+1:

Just in case to prevent future regressions, would you mind adding a test case in `cmd/mysqldef/mysqldef_test.go`?

PR #27, opened 2019-06-12
Fix column order.

kawakami-o3:
This PL emits `alter ... after...` even if only the order of the columns is different.


k0kubun:
Thank you! :+1:

kawakami-o3:
Wow, thanks.

PR #28, opened 2019-07-16
Fix to preserve line feeds when using stdin

YuukiARIA:
Currently `sqldef.readFile` function removes line ending characters when input from stdin.
DDL is almost valid even if their lines are combined, but line-sensitive syntax (e.g. `-- line comments`) is broken.

Here is an example schema, including simple table definition written in multiline.

```
CREATE TABLE `sample` (
  `id` INT -- ID
);
```

`--file` option works well because it preserves line endings.

```
% mysqldef -uroot example --dry-run --file sample.sql 
-- dry run --
CREATE TABLE `sample` (
  `id` INT -- ID
);
```

However stdin is not working.

```
% mysqldef -uroot example --dry-run < sample.sql
found syntax error when parsing DDL "CREATE TABLE `sample` (  `id` INT -- ID)": syntax error at position 41
```

Given lines are combined, and `-- ID` is no longer valid line comment.

I tested this with mac and mysql, but other platform or other db adapters are considered to have same issue.

### DIFF
--- a/sqldef.go
+++ b/sqldef.go
@@ -60,27 +60,24 @@ func Run(generatorMode schema.GeneratorMode, db adapter.Database, options *Optio
 }
 
 func readFile(filepath string) (string, error) {
-	var content string
 	var err error
+	var buf []byte
+
 	if filepath == "-" {
 		stat, _ := os.Stdin.Stat()
 		if (stat.Mode() & os.ModeCharDevice) != 0 {
 			return "", fmt.Errorf("stdin is not piped")
 		}
 
-		var buf []byte
 		buf, err = ioutil.ReadAll(os.Stdin)
-		content = string(buf)
 	} else {
-		var buf []byte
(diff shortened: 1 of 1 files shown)


k0kubun:
Thank you. I'll check this again a few days later as I'm going to a vacation now.

k0kubun:
To prevent the same bug from happening again, it would be nice to have a test case that reproduces the possible syntax error due to the bug when using stdin. However, having this fix sounds totally fine. So I'm merging your patch as is. Thank you!

k0kubun:
This will be released as v0.4.13 in [this Travis build](https://travis-ci.org/k0kubun/sqldef/builds/561075189).

YuukiARIA:
Thank you for merging!

> To prevent the same bug from happening again, it would be nice to have a test case

I totally agree. I'll try next chance! Thanks!

PR #31, opened 2019-08-27
Add support for BIGSERIAL datatype

yudppp:
I want to use it on postgressql. But not support `BIGSERIAL`.

I implement support BIGSERIAL.

Thank you.


yudppp:
Oh Sorry. I failed test.

Please waiting for resolve it :bow:

yudppp:
Umm, It was not easy for me.

I wrote this create create query

```sql
CREATE TABLE public.users (
    id BIGSERIAL NOT NULL,
    name TEXT
);
```

But pg_dump results is 

```sql
CREATE TABLE public.users (
    id bigint NOT NULL,
    name text
);
ALTER SEQUENCE public.users_id_seq OWNED BY public.users.id;
ALTER TABLE ONLY public.users ALTER COLUMN id SET DEFAULT nextval('public.users_id_seq'::regclass)
```

I checked pg_dump option, but does not change from separate queries.

Should I fix this following logic to `func (d *PostgresDatabase) DumpTableDDL(table string) (string, error)`?

https://stackoverflow.com/questions/33038651/pg-dump-serial-datatype-issues?answertab=active#tab-top

@k0kubun 
Do you have any good ideas?

k0kubun:
Hmm, that's a little complicated one like you said. 

> Should I fix this following logic to `func (d *PostgresDatabase) DumpTableDDL(table string) (string, error)` ?

That'd be one option.

Another option would be to parse `ALTER SEQUENCE` statement and calculate `BIGSERIAL` for the column in `convertDDLsToTables` https://github.com/k0kubun/sqldef/blob/v0.4.14/schema/generator.go#L540-L580.

yudppp:
thank you.
I see.
I will try fix `convertDDLsToTables` function this weekend.

Issue #10, opened 2018-09-03
Support ALTER COLUMN for PostgreSQL


k0kubun:
"ADD / DROP PRIMARY KEY" and "Drop pg_dump dependency" (#32) have been done, so for better PostgreSQL support, we would also like:

* Support ALTER COLUMN for PostgreSQL

PR #32, opened 2019-09-04
Remove pg_dump dependency

yudppp:
I checked `Drop pg_dump dependency` issue. https://github.com/k0kubun/sqldef/issues/10.
Easy to implement `Drop pg_dump dependency` and `Support SERIAL` together.

@k0kubun 
Please check my plan is correct.

## TODO
- Support SERIAL -> I will fix other PR(this PR is `Drop pg_dump dependenc` only)
- [x] Add test code
- [x] Support multi primary key

### DIFF
--- a/adapter/postgres/postgres.go
+++ b/adapter/postgres/postgres.go
@@ -4,13 +4,14 @@ import (
 	"database/sql"
 	"fmt"
 	"os"
-	"os/exec"
-	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/k0kubun/sqldef/adapter"
 	_ "github.com/lib/pq"
 )
+
+const indent = "    "
 
 type PostgresDatabase struct {
 	config adapter.Config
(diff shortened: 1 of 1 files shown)


k0kubun:
Wow, that looks great. I think it's okay to go with this once the existing integration tests pass. I appreciate your effort working on this.

yudppp:
@k0kubun 
I Finished.And passed all existing integration tests

k0kubun:
Sounds good. Thanks a lot!

adapter/postgres/postgres.go
```diff
@@ -47,103 +48,241 @@ func (d *PostgresDatabase) TableNames() ([]string, error) {
...
+			}
+		}
+		col.Name = strings.Trim(colName, `" `)
+		if colDefault != nil || isPK {
+			if isPK {
+				col.IsPrimaryKey = true
+			} else {
+				col.Default = *colDefault
+			}
+		}
```
k0kubun, line 174: Just curious, did you have any reason why this wasn't:

```go
		if isPK {
			col.IsPrimaryKey = true
		}
		if colDefault != nil {
			col.Default = *colDefault
		}
```
More specifically, did you intend to skip setting `col.Default` when `colDefault != nil` and `isPk` are true?

PR #33, opened 2019-09-06
Support some serial for postgresql

yudppp:
Support smallserial, serial and bigserial.


cmd/psqldef/psqldef_test.go
```diff
@@ -234,11 +233,8 @@ func TestPsqldefDataTypes(t *testing.T) {
 	//   - polygon
 	//   - real
 	//   - float4
-	//   - smallint
```
yudppp: `smallint` is already support.

PR #41, opened 2019-10-01
Support Partial Indexes syntax(PostgreSQL) 

yudppp:
Support Partial Indexes syntax.

example
```sql
CREATE UNIQUE INDEX index_name on users (name) WHERE age > 20;
```

https://www.postgresql.org/docs/9.4/indexes-partial.html


schema/generator.go
```diff
@@ -657,6 +657,9 @@ func areSameIndexes(indexA Index, indexB Index) bool {
 			return false
 		}
 	}
+	if indexA.where != indexB.where {
```
yudppp: This expression is not thorough.
For example, `age > 20` and `20 < age` deal with different expression.(Should be like same)

Should I support this case?

k0kubun: It'd be nice for have for sure, but I think it's not critical because it's already usable in a normal way (if we don't attempt refactoring, etc.) and we can add the support later when we want.

Issue #43, opened 2019-10-27
I want an option that doesn't make disruptive changes.


touyu:
When developing with multiple people separately, if migration is performed for each, the table will be dropped.
Therefore I want an option that doesn't make destructive changes like Gorm's automigrate.

k0kubun:
I'm basically open to merge https://github.com/k0kubun/sqldef/pull/44, but I'm not sure if I understand your situation correctly. 

I guess the situation is like:

1. There was a table "T" on master branch.
2. On a topic branch, a person "A" added a column "a" to the table "T".
3. On another topic branch, a person "B" added a column "b" to the table "T".

In this situation, step 3 would unintentionally drop the column "a". Therefore you'd like to use the `--safety` option on step 2 or 3, and you'd run schema migrations without `--safety` only when you really want to drop a column. Is my understanding correct?

touyu:
Yes, your understanding is correct.
I assume the similar situation.

PR #44, opened 2019-10-27
[Proposal] Add --skip-drop option

touyu:
issue: https://github.com/k0kubun/sqldef/issues/43

I implemented the `--skip-drop` option as one way to avoid destructive changes.
Please consider it.

### DIFF
--- a/adapter/database.go
+++ b/adapter/database.go
@@ -50,7 +50,7 @@ func RunDDLs(d Database, ddls []string, isSafety bool) error {
 	fmt.Println("-- Apply --")
 	for _, ddl := range ddls {
 		if isSafety && strings.Contains(ddl, "DROP") {
-			fmt.Printf("Not executed: %s", ddl)
+			fmt.Printf("Not executed: %s;\n", ddl)
 			continue
 		}
 		fmt.Printf("%s;\n", ddl)
--- a/adapter/database.go
+++ b/adapter/database.go
@@ -50,7 +50,7 @@ func RunDDLs(d Database, ddls []string, skipDrop bool) error {
 	fmt.Println("-- Apply --")
 	for _, ddl := range ddls {
 		if skipDrop && strings.Contains(ddl, "DROP") {
-			fmt.Printf("Not executed: %s;\n", ddl)
+			fmt.Printf("-- Skipped: %s;\n", ddl)
 			continue
 		}
 		fmt.Printf("%s;\n", ddl)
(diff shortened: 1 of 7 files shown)


cmd/mysqldef/mysqldef.go
```diff
@@ -26,6 +26,7 @@ func parseOptions(args []string) (adapter.Config, *sqldef.Options) {
 		File     string `long:"file" description:"Read schema SQL from the file, rather than stdin" value-name:"sql_file" default:"-"`
 		DryRun   bool   `long:"dry-run" description:"Don't run DDLs but just show them"`
 		Export   bool   `long:"export" description:"Just dump the current schema to stdout"`
+		Safety   bool   `long:"safety" description:"Not make destructive changes such as Drop"`
```
k0kubun: I feel `--skip-drop` might be more descriptive about what it does than `--safety`. WDYT?

adapter/database.go
```diff
@@ -42,13 +42,17 @@ func DumpDDLs(d Database) (string, error) {
...
-func RunDDLs(d Database, ddls []string) error {
+func RunDDLs(d Database, ddls []string, isSafety bool) error {
 	transaction, err := d.DB().Begin()
 	if err != nil {
 		return err
 	}
 	fmt.Println("-- Apply --")
 	for _, ddl := range ddls {
+		if isSafety && strings.Contains(ddl, "DROP") {
+			fmt.Printf("Not executed: %s;\n", ddl)
```
k0kubun: sqldef's output is designed to always be a valid input for `mysql`/`psql`. Could you prefix `--` to make it a SQL comment?

And if you follow https://github.com/k0kubun/sqldef/pull/44/files#r339372757, it'd be `-- Skipped: %s;`.

k0kubun:
Could you add the same option to psqldef for consistency, and update the README as well?

Also, it'd be really nice if we can have tests in mysqldef_test.go and psqldef_test.go.

cmd/mysqldef/mysqldef.go
touyu: I also think `--skip-drop` is better than `--safety` !

adapter/database.go
touyu: I understand.
I think `-Skipped:% s;` is good.

touyu:
@k0kubun 
Thank you for reviews.
I would be happy if you could check it again.

Issue #68, opened 2020-10-21
CHARACTER SET and COLLATE should be placed before [NOT NULL | NULL] on MySQL


hottestseason:
## Problem

SSIA

https://dev.mysql.com/doc/refman/5.6/en/create-table.html
https://dev.mysql.com/doc/refman/5.6/en/string-type-syntax.html

## Expected

```sql
ALTER TABLE users ADD COLUMN name varchar(20) CHARACTER SET ascii COLLATE ascii_bin NOT NULL AFTER id;
```

### Initial schema

```sql
CREATE TABLE users (
  id bigint NOT NULL,
  PRIMARY KEY (id)
);
```

### Input DDLs

```sql
CREATE TABLE users (
  id bigint NOT NULL,
  name VARCHAR(20) CHARACTER SET ascii COLLATE ascii_bin NOT NULL,
  PRIMARY KEY (id)
);
```

### Output DDLs

```sql
ALTER TABLE users ADD COLUMN name varchar(20) NOT NULL CHARACTER SET ascii COLLATE ascii_bin AFTER id;
```

k0kubun:
Merged your PR, thanks!

PR #69, opened 2020-10-21
Fix charset and not null order on MySQL

hottestseason:
refs: #68

I fix this very naively but it is great if there are more smart ways


k0kubun:
Thanks for reporting and fixing the issue!

PR #77, opened 2020-11-22
feat: Suport CREATE POLICY (for psqldef)

jumpeiMano:
I have made additional implementations to support the following features:
- [POLICY](https://www.postgresql.org/docs/current/sql-createpolicy.html) (for psqldef)

And I also fix a bug of psqldef. When I used `ALTER TABLE ONLY` statements to set PRIMARY KEY, I received some errors like this.
```
unexpected ddl type in generateDDLs: &{ALTER TABLE ONLY dbname.tablename
    ADD CONSTRAINT pkname PRIMARY KEY (columnname) dbname.tablename {pkname  [{columnname <nil>}] false false }}
exit status 1
```
So I fixed it.

PR #78, opened 2020-11-22
feat: support CREATE (OR REPLACE) VIEW

jumpeiMano:
This PR is a split from https://github.com/k0kubun/sqldef/pull/77.

I have made additional implementations to support the following features:
- VIEW
  - CREATE VIEW, DROP VIEW (for all mode)
  - CREATE OR REPLACE VIEW (for mysqldef and psqldef)


adapter/mysql/mysql.go
```diff
@@ -56,6 +56,28 @@ func (d *MysqlDatabase) DumpTableDDL(table string) (string, error) {
...
+	for rows.Next() {
+		var viewName, viewType, definition string
+		if err = rows.Scan(&viewName, &viewType); err != nil {
+			return nil, err
+		}
+		query := fmt.Sprintf("select VIEW_DEFINITION from INFORMATION_SCHEMA.VIEWS where TABLE_NAME = '%s';", viewName)
+		if err = d.db.QueryRow(query).Scan(&definition); err != nil {
+			return nil, err
+		}
+		ddls = append(ddls, fmt.Sprintf("CREATE OR REPLACE VIEW %s AS %s", viewName, definition))
```
k0kubun, line 76: I assume this means users need to specify `CREATE OR REPLACE VIEW` in a schema definition to satisfy `strings.ToLower(currentView.definition) == strings.ToLower(desiredView.definition)`.

But I think `OR REPLACE` sounds like an implementation detail to me, and because other resources like tables are (at least documented to be) declared as `CREATE TABLE`, I think VIEW should be declared with `CREATE VIEW` without OR REPLACE for consistency across sqldef.

Or do you have any motivation to prefer `CREATE OR REPLACE VIEW` in a schema file rather than `CREATE VIEW`? Ideally sqldef should support both `CREATE VIEW` and `CREATE OR REPLACE VIEW`, but if we were to choose one, I'd like to choose `CREATE VIEW`.

jumpeiMano: Thank you for your reviewing! i fixed it!
https://github.com/k0kubun/sqldef/pull/78/commits/c5287ad24736e8df22414dc96b13202141c58688

k0kubun: Oops, I was misunderstanding your implementation. I thought `currentView.definition` includes the `CREATE OR REPLACE VIEW`, but it wasn't. So this part didn't really matter. Anyway the change makes sense for similar reasons, thanks!

README.md
```diff
@@ -352,6 +370,21 @@ Remove the line to DROP INDEX.
...
+
+```diff
+ CREATE OR REPLACE VIEW foo AS
+   select u.id as id, p.id as post_id
+   from  (
+     mysqldef_test.users as u
+     join mysqldef_test.posts as p on ((u.id = p.user_id))
+   )
+ ;
++ CREATE OR REPLACE VIEW foo AS select u.id as id, p.id as post_id from (mysqldef_test.users as u join mysqldef_test.posts as p on (((u.id = p.user_id) and (p.is_deleted = 0))));
```
k0kubun, line 383:
```suggestion
+ CREATE OR REPLACE VIEW foo AS select u.id as id, p.id as post_id from (psqldef_test.users as u join psqldef_test.posts as p on (((u.id = p.user_id) and (p.is_deleted = 0))));
```

because this is an example for PostgreSQL.

README.md
```diff
@@ -303,6 +306,21 @@ Remove the line to DROP FOREIGN KEY.
...
 Composite foreign key may not work for now.
 
+### CREATE (OR REPLACE) VIEW
+
+```diff
+ CREATE OR REPLACE VIEW foo AS
+   select u.id as id, p.id as post_id
+   from  (
+     mysqldef_test.users as u
+     join mysqldef_test.posts as p on ((u.id = p.user_id))
```
k0kubun, line 316: Is that intentional that a view definition needs to include a database name, or is that a compromise for making the current implementation work? Not complaining about that in this PR, but I'd like to check what you'd actually like for future changes.

README.md
k0kubun: Or, according to the tests, it seems a database name is not needed for PostgreSQL.
```suggestion
+ CREATE OR REPLACE VIEW foo AS select u.id as id, p.id as post_id from (users as u join posts as p on (((u.id = p.user_id) and (p.is_deleted = 0))));
```

adapter/mysql/mysql.go
k0kubun: Since you added the commit, let me be consistent everywhere https://github.com/k0kubun/sqldef/pull/78/commits/ec553382cbbdf49dbe06d916ef7d7dd70b192350.

k0kubun (approved):
Thank you :+1:

README.md
jumpeiMano: Oh... I'm so sorry about that 🙇

README.md
jumpeiMano: The result of [this query](https://github.com/k0kubun/sqldef/blob/e94187cf27/adapter/mysql/mysql.go#L60) it is included a database name.

k0kubun: Yes, I know it, and that's not what I'm asking. "Is that a compromise for making the current implementation (using `show full tables`) work?"

jumpeiMano: I'm sorry, I'm not good at English. The answer is "yes".

PR #80, opened 2020-12-01
feat: support variety of the default value in psqldef

jumpeiMano:
This PR is a split from https://github.com/k0kubun/sqldef/pull/77.

I have made additional implementations to support the following features:

- default values (psqldef)
  - `xxx character varying(50) NOT NULL DEFAULT ''::character varying`
  - `yyy_at timestamp with time zone NOT NULL DEFAULT '0001-01-01 00:00:00'::timestamp without time zone`
  - `zzz_at timestamp with time zone DEFAULT now()`


k0kubun (approved):
Thank you!

PR #81, opened 2020-12-02
feat: support SQLite AUTOINCREMENT column attribute

tsuty:
I implemented for 'AUTOINCREMENT' column attribute syntax for SQLite. 

I thought that 'ColumnType' should need another `Autoincrement` field for sqlite, or 'ParserMode'.
This implementation is the latter.


sqlparser/parser.y
```diff
@@ -789,6 +789,12 @@ column_definition_type:
     $1.Autoincrement = BoolVal(true)
     $$ = $1
   }
+| column_definition_type AUTOINCREMENT
+  {
+    $1.Autoincrement = BoolVal(true)
+    $1.ParserMode = ParserModeSQLite3
```
k0kubun, line 795: `ParserMode` doesn't feel like an attribute of ColumnType. WDYT about adding `AutoIncrement`, using it for `AUTO_INCREMENT`, and using `Autoincrement` for `AUTOINCREMENT` instead?

k0kubun: Hmm, but then other implementations would need to deal with both. Let me reconsider it.

k0kubun:
> I thought that 'ColumnType' should need another Autoincrement field for sqlite, or 'ParserMode'.
This implementation is the latter.

This seems to be used only for `*ColumnType.Format()`. Is the `Format()` method used by sqldef? If not, we might be able to simply skip both of them.

Also, it'd be nice if you can add tests to `cmd/sqlite3def/sqlite3def_test.go` instead. We don't really maintain `sqlparser/parse_test.go`. If the test is the only reason to maintain the `ParserMode`, it's fine to just remove the tests.

tsuty:
> This seems to be used only for *ColumnType.Format(). Is the Format() method used by sqldef? If not, we might be able to simply skip both of them.

sqldef does not use it. 

> We don't really maintain sqlparser/parse_test.go If the test is the only reason to maintain the ParserMode, it's fine to just remove the tests.

I understand. implement it.
I am only aiming for no syntax error for sqlite3def command.
And I just noticed that need implement for sqlite internal tables 
 (sqlite_sequence...) 

So, I need more  imprements.

k0kubun:
Please let me know when your next patch becomes ready. I'm closing this for now since I can't merge this PR due to conflicts.

PR #86, opened 2020-12-12
Add `t.Helper()`

178inaba:
When the test is in error, the number of lines of the test helper is output as shown below.

```
sqlite3def_test.go:212: failed to execute 'sqlite3def sqlite3def_test --file schema.sql' (error: 'exit status 1'): `found syntax error when parsing DDL "CREATE VIEW `view_users` AS select id from users where age = 1": syntax error at position 14 near '`'
```

This makes it difficult to investigate errors, so use `t.Helper()` to display the number of lines called.
https://golang.org/pkg/testing/#T.Helper


k0kubun (approved):
Thanks! :+1:

PR #87, opened 2020-12-12
Add identifier escaping process

178inaba:
If the identifier is not escaped, an error will occur when handling the same table name or column name as the reserved word.
So I added a process to escape the identifier.

ref.
https://github.com/k0kubun/sqldef/commit/7e4a2dd53116551e9cb1d7e57f5d13a962c08e89


k0kubun (approved):
Perfect, thank you so much!

PR #97, opened 2021-02-26
feat: support 'CHECK' option of CREATE TABLE command for psqldef

jumpeiMano:
I have made additional implementations to support `CHECK` option of CREATE TABLE command in psqldef.
For more information about `CHECK` option, please see this documentation. https://www.postgresql.org/docs/current/sql-createtable.html


k0kubun (approved):
Thanks!

PR #104, opened 2021-04-08
context.Context support, updated to go1.16

johejo:


cmd/psqldef/psqldef.go
```diff
@@ -104,5 +106,8 @@ func main() {
 	}
 	defer database.Close()
 
-	sqldef.Run(schema.GeneratorModePostgres, database, options)
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	defer cancel()
```
johejo, line 110: Is it OK to propagate context cancel  by syscall?
I would like to ask the maintainer's intention regarding the use case of canceling the migration.

k0kubun:
Could you explain how you're going to use this change? Do you intend to use sqldef as a library? (It's not supported)

johejo:
Handle the signal appropriately and cancel the execution of the query.
This is to make the migration safe.
I am not thinking of using sqldef as a library.

johejo:
My concern may be off the mark. 
Do we get the same result if canceling the ExecContext or if terminating the process with SIGINT or SIGTERM?

k0kubun:
> Handle the signal appropriately and cancel the execution of the query.
> This is to make the migration safe.
> I am not thinking of using sqldef as a library.

That makes sense. It was unclear from "context.Context support, updated to go1.16" and the blank description because TBH I'm not really familiar with `context.Context`.

> My concern may be off the mark.
> Do we get the same result if canceling the ExecContext or if terminating the process with SIGINT or SIGTERM during Exec?
>
> How should sqldef behave during interrupt shutdown?

Whether we use `context.Context` or not, this discussion is nice to have. Thank you for bringing this up.

May I ask what made you realize improve this thing? Did you see any unexpected result by SIGINT/SIGTERM, or can you think of a scenario that can be improved this PR?

johejo:
As you pointed out, my PR description was inadequate, sorry.
Anyway, the upgrade to Go 1.16 will be separated into a separate PR.

johejo:
Rebuild PRs.

PR #107, opened 2021-04-09
Update to go1.16 and drop darwin/386 support

johejo:
darwin/386 is no longer supported since go1.15.
https://github.com/golang/go/issues/37610

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,6 @@ package:
 	$(MAKE) package-targz GOOS=linux GOARCH=arm64
 	$(MAKE) package-targz GOOS=linux GOARCH=arm
 	$(MAKE) package-zip GOOS=darwin GOARCH=amd64
-	$(MAKE) package-zip GOOS=darwin GOARCH=386
 	$(MAKE) package-zip GOOS=windows GOARCH=amd64
 	$(MAKE) package-zip GOOS=windows GOARCH=386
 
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -5,16 +5,12 @@ services:
...
-      - ./docker/mysql/data:/var/lib/mysql
     ports:
       - '3306:3306'
   postgres:
     image: postgres:12
     environment:
       POSTGRES_USER: postgres
       POSTGRES_HOST_AUTH_METHOD: trust
-    volumes:
-      - ./docker/postgresql/data:/var/lib/postgresql/data
```
k0kubun, line 18: Why did you remove it?

Makefile
```diff
@@ -27,7 +27,6 @@ package:
 	$(MAKE) package-targz GOOS=linux GOARCH=arm64
 	$(MAKE) package-targz GOOS=linux GOARCH=arm
 	$(MAKE) package-zip GOOS=darwin GOARCH=amd64
-	$(MAKE) package-zip GOOS=darwin GOARCH=386
```
k0kubun, line 30: Did Go stop supporting it?

docker-compose.yml
johejo: `go get -t ./...` fails in github actions
https://github.com/k0kubun/sqldef/runs/2302532462?check_suite_focus=true

Makefile
johejo: See PR description(edited)

k0kubun (approved):

PR #114, opened 2021-05-17
Support PostgreSQL Identity Column

tsuty:
PostgreSQL `GENERATED ALWAYS AS IDENTITY` and `GENERATED BY DEFAULT AS IDENTITY` supports

This code can parse the sequence options basically.(like START WITH 10)
But I couldn't completely build sequence options from current schema. 
I'm lost. :tired_face:
Hmm ...
I don't implement changing sequence options. :bow:

I feel the changing sequence options is different from changing DDL. Um ....
I think There are so few cases. :sweat_smile:


k0kubun:
Thanks!

> But I couldn't completely build sequence options from current schema.

Sorry, I didn't get what you mean by this. Can you write a test case where the current implementation is problematic?

> I don't implement changing sequence options. bow
>
> I feel the changing sequence options is different from changing DDL. Um ....
> I think There are so few cases.

It seems fine. You don't need to implement what you don't use.

tsuty:
> > But I couldn't completely build sequence options from current schema.

> Sorry, I didn't get what you mean by this. Can you write a test case where the current implementation is problematic?

Sorry I will explain about that.

We can know sequence option from `information_schema.columns`.
But, I don't fully judge that option by DDL or default.

```
postgres=# create table entries
postgres-# (
postgres(#   id bigint GENERATED ALWAYS AS IDENTITY,
postgres(#   name varchar(128) NOT NULL
postgres(# );
CREATE TABLE
postgres=# select
postgres-#   table_name
postgres-#   , column_name
postgres-#   , is_identity
postgres-#   , identity_generation
postgres-#   , identity_start
postgres-#   , identity_increment
postgres-#   , identity_maximum
postgres-#   , identity_minimum
postgres-#   , identity_cycle
postgres-# from information_schema.columns
postgres-# where table_name = 'entries' and column_name = 'id'\x
Expanded display is on.
postgres-# ;
-[ RECORD 1 ]-------+--------------------
table_name          | entries
column_name         | id
is_identity         | YES
identity_generation | ALWAYS
identity_start      | 1
identity_increment  | 1
identity_maximum    | 9223372036854775807
identity_minimum    | 1
identity_cycle      | NO
```

`pg_dump` too.   I only defined `GENERATED ALWAYS AS IDENTITY`.
 
```
ALTER TABLE public.entries ALTER COLUMN id ADD GENERATED ALWAYS AS IDENTITY (
    SEQUENCE NAME public.entries_id_seq
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1
);
```

Am I making sense?

k0kubun:
Looking at the code, it seems that:

* `buildDumpTableDDL` can generate `GENERATED %s AS IDENTITY` as a part of a `CREATE TABLE`.
  * The `%s` supports what `information_schema.columns.identity_generation` returns (only `ALWAYS` and `BY DEFAULT` are parsable).
  * Ideally all other `information_schema.columns.identity_*` should be supported, but it's not supported for now. (I guess this is what you pointed out?)
* parser.y supports `'(' sequence_opt ')'` in addition to what `buildDumpTableDDL` can generate
  * So, while `hasSameIdentity` checks only `IdentityOpt.Behavior`, you could specify `sequence_opt` which parser.y can parse and the generator generates `sequence_opt` when the identity is newly added or `Behavior` is changed.

Did I get it right? The implementation seems reasonable.

schema/parser.go
```diff
@@ -373,3 +374,53 @@ func detectCharset(table sqlparser.TableSpec) string {
...
+}
+
+func parseIdentityOpt(opt *sqlparser.IdentityOpt) *IdentityOpt {
+	if opt == nil {
+		return nil
+	}
+	dest := &IdentityOpt{
+		Behavior: strings.ToUpper(opt.Behavior),
+	}
+	if opt.Sequence != nil {
```
k0kubun, line 390: I think building the `Sequence` string from various options is responsibility of generator.go. Could you change the type of `schema.IdentityOpt.Sequence` from `string` to a `schema.Sequence` struct that looks similar to `sqlparser.Sequence` (but with simpler field types), and move this logic to generator.go?

tsuty:
> The implementation seems reasonable.

Thank you. Sorry for my poor explanation.

> should be supported, but it's not supported for now.

> the identity is newly added or Behavior is changed.

That's right 

The sequence of identity column option applies only 2 ways.
1. When creating a table. 
3. When change to identity column from non identity column.(adding too)

Not support: changing sequence from already applied identity column.

schema/parser.go
tsuty: changed ! 3ace2b11a2aa551c8f2519f7f2ac27b15244afdc

k0kubun (approved):
Thanks!

PR #115, opened 2021-05-19
Support more default value types

tsuty:
This is supporting more default value types for postgres.

- CURRENT_DATE
- CURRENT_TIME, CURRENT_TIME(d)
- text 
- bpchar ( `CHARACTER(n), CHAR(n)` fixed-length character)

Not yet supported for all types


k0kubun (approved):
Thanks :+1:

PR #116, opened 2021-05-19
Support changing default

tsuty:
This is supporting changing default  for postgres.


k0kubun (approved):
This is a great addition. Thank you!

PR #117, opened 2021-05-20
fix postgres environment variable

tsuty:
I don't know if it's intentional implements that supporting `PGPASS` environment variable to set password on psqldef

The psql command used `PGPASSWORD` environment variable.

https://www.postgresql.org/docs/current/libpq-envars.html
> PGPASSWORD behaves the same as the password connection parameter. 

Should I support `PGPASS` for backward compatibility?

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Usage:
 
 Application Options:
   -U, --user=username        PostgreSQL user name (default: postgres)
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
What a coincidence, I tried to use `$PGPASS` today and yes, only `$PGPASSWORD` worked. I think we don't need to support `$PGPASS`.

PR #121, opened 2021-06-10
mssqldef: add some feature

ytakaya:
I added following features to mssqldef.
- Column: ADD COLUMN, DROP COLUMN
- Primary Key: ADD PRIMARY KEY, DROP PRIMARY KEY

I ran into a problem when supporting `DROP COLUMN`.
MSSQL does not allow to drop columns with default or primary key constraints.
In addition, in order to delete a specific constraint, we need the constraint name defined on DB.
What I mean is that I cannot delete in the following way.
```
ALTER TABLE [table_name] ALTER [column_name] DROP DEFAULT
```
I need to use the following way.
```
ALTER TABLE [table_name] DROP CONSTRAINT [DF_constraint_name]
```
Similarly, we need to use the constraint name in the drop of primary key.
So I needed to extract the constraint names from the DB in some way and use it in generator.
Finally, I implemented it to extract the constraint definition of the table in the adapter.
I tried to implement it in such a way that it would not affect other DBs, but since I made some changes to the interface and struct, I would like you to point out any concerns.


sqldef.go
```diff
@@ -20,7 +20,7 @@ type Options struct {
 
 // Main function shared by `mysqldef` and `psqldef`
 func Run(generatorMode schema.GeneratorMode, db adapter.Database, options *Options) {
-	currentDDLs, err := adapter.DumpDDLs(db)
+	currentDDLs, constraints, err := adapter.DumpDDLs(db)
```
k0kubun, line 23: The current design is to compare bare string DDLs and produce DDLs, instead of returning well-structured data from `DumpDDLs`. 

Could you please try not to change existing APIs used by other adapters? At least in this place, I thought you could produce DDLs for the constraints and let the parser parse them. That allows us to symmetrically implement the parsed results and make it harder to introduce inconsistency on one side.

adapter/mysql/mysql.go
```diff
@@ -78,6 +78,10 @@ func (d *MysqlDatabase) Views() ([]string, error) {
 	return ddls, nil
 }
 
+func (d *MysqlDatabase) Constraints(table string) ([]*adapter.ColumnConstraints, error) {
```
k0kubun, line 81: If you followed the current design https://github.com/k0kubun/sqldef/pull/121#discussion_r649057839, I don't think you needed to add this API to other adapters.

k0kubun:
May I get each of the following features in different PRs instead of as one PR for "some feature"?

> * Column: ADD COLUMN, DROP COLUMN
> * Primary Key: ADD PRIMARY KEY, DROP PRIMARY KEY

Currently, it's harder for reviewers to understand which part is needed for what purpose.

sqldef.go
ytakaya: Thanks for the quick reply.
> At least in this place, I thought you could produce DDLs for the constraints and let the parser parse them.

It certainly looks like I can include existing constraint names in the DDLs.

I'll try to come up with an implementation that doesn't change the current structure.

ytakaya:
Thank you for the quick response.
I'll take your advice and rethink how to implement it. I will then split this pull request and send in a unit that is easy to understand the changes.

ytakaya:
Can I close this pull request because I'm going to propose adding a feature in another pull request with a different policy?

k0kubun:
Let me close this since it's not actionable from my side for now. If you still need something else here, please rebase and reopen this or open another PR.

PR #125, opened 2021-06-18
mssqldef: support index option

ytakaya:
This changes is to support index option of mssql.
I am using this change to parse indexes with options, not to compare indexes including the value of the options.

`mssql_index_option_list` is needed to avoid `shift/reduce conflict`, which occurs when I write a description using `,` for an existing `index_option_list`.


k0kubun (approved):
:+1:

PR #126, opened 2021-06-21
mssqldef: support index

ytakaya:
This change supports the following features.
- ADD INDEX
- DROP INDEX

### DIFF
--- a/schema/generator.go
+++ b/schema/generator.go
@@ -398,7 +398,7 @@ func (g *Generator) generateDDLsForCreateTable(currentTable Table, desired Creat
 			// Drop and add index as needed.
 			switch g.mode {
 			case GeneratorModeMssql:
-				if !areSameIndexes(*currentIndex, desiredIndex) || !areSameIndexOptions(currentIndex.options, desiredIndex.options) {
+				if !areSameIndexes(*currentIndex, desiredIndex) {
 					ddls = append(ddls, g.generateDropIndex(desired.table.name, desiredIndex.name))
 					ddls = append(ddls, g.generateDDLForCreateIndex(desired.table.name, desiredIndex))
 				}
@@ -803,6 +803,18 @@ func (g *Generator) generateIndexDefinition(index Index) string {
 			strings.Join(columns, ", "),
 		)
 	}
+
+	if len(index.options) > 0 {
+		switch g.mode {
+		case GeneratorModeMysql:
+			indexOption := index.options[0]
+			if indexOption.optionName == "parser" {
+				indexOption.optionName = "WITH " + indexOption.optionName
+			}
+			definition += fmt.Sprintf(" %s %s", indexOption.optionName, string(indexOption.value.raw))
+		}
+	}
+
 	return definition
 }
 
@@ -1211,13 +1223,9 @@ func areSameIndexes(indexA Index, indexB Index) bool {
 		return false
 	}
 
-	return true
-}
-
-func areSameIndexOptions(currentOptions []IndexOption, desiredOptions []IndexOption) bool {
-	for _, desiredOption := range desiredOptions {
-		if currentOption := findIndexOptionByName(currentOptions, desiredOption.optionName); currentOption != nil {
-			if !areSameValue(currentOption.value, desiredOption.value) {
+	for _, optionB := range indexB.options {
+		if optionA := findIndexOptionByName(indexA.options, optionB.optionName); optionA != nil {
+			if !areSameValue(optionA.value, optionB.value) {
 				return false
 			}
 		} else {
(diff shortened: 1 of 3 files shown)


schema/generator.go
```diff
@@ -396,13 +396,26 @@ func (g *Generator) generateDDLsForCreateTable(currentTable Table, desired Creat
 
 		if currentIndex := findIndexByName(currentTable.indexes, desiredIndex.name); currentIndex != nil {
 			// Drop and add index as needed.
-			if !areSameIndexes(*currentIndex, desiredIndex) {
-				ddls = append(ddls, fmt.Sprintf("ALTER TABLE %s DROP INDEX %s", g.escapeTableName(desired.table.name), g.escapeSQLName(currentIndex.name)))
-				ddls = append(ddls, fmt.Sprintf("ALTER TABLE %s ADD %s", g.escapeTableName(desired.table.name), g.generateIndexDefinition(desiredIndex)))
+			switch g.mode {
+			case GeneratorModeMssql:
```
k0kubun, line 400: This is applicable to all of your patches, but please consider NOT to introduce `g.mode` to a place where `g.mode` switch is not used. At least I couldn't understand why you needed to exclude `areSameIndexOptions` from `areSameIndexes`, and using a name like `generateDDLForCreateIndex` only for mssql is pretty confusing while you already have `generateIndexDefinition` for the same purpose.

k0kubun (changes requested):
May I see your effort to reduce the number of new branches for `GeneratorModeMssql` first? Feel free to modify existing implementations for doing so.

I'm not saying you must not use `g.mode` switch everywhere. The general idea is that you should use `g.mode` switches preferably in deeply nested function calls rather than more abstract functions that already do a lot more things (e.g. `generateDDLsForCreateTable`). For example, if you generate entire DDLs from `generateIndexDefinition`, you could move a `g.mode` switch from `generateDDLsForCreateTable` to `generateIndexDefinition`.

The reason why I'm suggesting these changes is that first of all, ideally, every `g.mode` switch should be eliminated. But if you need to introduce it, you should consider introducing it to a trivial location that doesn't really impact an overall code flow or readability. As I said in the above comment, your current patch duplicates a function with similar roles, branches them in a relatively abstract place, and makes it hard to understand what is used by which mode.

ytakaya:
I apologize for the lack of explanation of the changes I made.
The following are the reasons why I needed `g.mode`.

In mssql, I couldn't use the syntax `ALTER TABLE` to put an index on an existing table, I had to use the syntax `CREATE INDEX`.
My addition of `generateDDLForCreateTable` is a function for the syntax `CREATE TABLE`. I made it a separate function because I thought its intent was different from the existing `generateIndexDefinition` for `ALTER TABLE`.
https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017#syntax-for-disk-based-tables

As for `areSameIndexOptions`, it was a branch born of my misunderstanding.
The syntax for the INDEX option existed in the existing parser.y, but it was not used in the generator.
Therefore, I mistakenly thought that DBs other than mssql were intentionally designed not to compare index options.

For these reasons, I thought that `areSameIndexOptions` should be removed, but what do you think should be done about `generateDDLsForCreateTable`?

k0kubun:
I knew you need `g.mode` for `ALTER TABLE`/`CREATE INDEX`. I'm just talking about where you use it. You should be able to have only one of `generateDDLForCreateIndex` and `generateIndexDefinition` and let it switch `ALTER TABLE` or `CREATE INDEX` using `g.mode` in it.

ytakaya:
I apologize for the lack of explanation of the changes I made. The following are the reasons why I needed g.mode.

In mssql, I couldn't use the syntax ALTER TABLE to put an index on an existing table, I had to use the syntax CREATE INDEX. My addition of generateDDLForCreateTable is a function for the syntax CREATE TABLE. I made it a separate function because I thought its intent was different from the existing generateIndexDefinition for ALTER TABLE. https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017#syntax-for-disk-based-tables

As for areSameIndexOptions, it was a branch born of my misunderstanding. The syntax for the INDEX option existed in the existing parser.y, but it was not used in the generator. Therefore, I mistakenly thought that DBs other than mssql were intentionally designed not to compare index options.

As you pointed out, I think that `areSameIndexOptions` should be removed, and for `generateDDLsForCreateTable`, putting the branch in an abstract place was not an optimal change.
Thank you for pointing this out, I'll try to fix it to improve the readability of the code.

k0kubun:
Not sure why you re-posted the comment, but https://github.com/k0kubun/sqldef/pull/126#issuecomment-865440350 was the answer to it.

ytakaya:
I changed the way to make a `generateAddIndex` and use `g.mode` in it to branch.

k0kubun (approved):
:+1: Thanks

PR #128, opened 2021-06-28
mssqldef: support check

ytakaya:
This change support check constraint.

As with the implementation of the default constraint, the constraint name had to be used for dropping the check constraint, so it was necessary to have the constraint name in the check struct.

The DROP and ADD parts of the check constraint are handled similarly to postgres, so I tried to make them common outside of the switch.
However, in that way, I would end up having to create a new branch outside the switch. I felt that it would be better to add a branch for mssql to the current switch, so I finally settled on the current form.


k0kubun (approved):
Thanks!

PR #130, opened 2021-07-02
psqldef: Escape DSN parts & Add PGSSLROOTCERT support

ackintosh:
### Escape DSN parts

Fixed the error on the DSN contains symbols.

```bash
$ export PGPASSWORD=".:?-&/%="
$ psqldef -U "user_.:?-&/%=" sandbox_db --skip-drop --file=test.sql
2021/07/02 16:26:34 Error on DumpDDLs: dial tcp: lookup user_.: no such host
```

### Add PGSSLROOTCERT support

https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT

I believe both of them are useful in case of AWS IAM Database Authentication. 🚀

### DIFF
--- a/adapter/postgres/postgres.go
+++ b/adapter/postgres/postgres.go
@@ -418,13 +418,17 @@ func postgresBuildDSN(config adapter.Config) string {
 		host = config.Socket
 	}
 
-	options := ""
+	var options []string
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
Thanks!

PR #131, opened 2021-07-02
mssqldef: Support more index options

ytakaya:
This added options that can be used in syntax related to index.

- `ASC` and `DESC`
  - Rebuild the index if the index direction changes.
  - If the adapter supports index direction, it can be used in other DBs.
- Extending the `CREATE INDEX` syntax
  - index option
    - This syntax can also be used in other DBs
  - `CLUSTERED` and `NONCLUSTERED`
  - `INCLUDE (column_name, ...)`
  - `ON  ...` (index partition)


schema/generator.go
```diff
@@ -787,6 +787,9 @@ func (g *Generator) generateAddIndex(table string, index Index) string {
 		if indexColumn.length != nil {
 			column += fmt.Sprintf("(%d)", *indexColumn.length)
 		}
+		if indexColumn.direction == "desc" {
```
k0kubun, line 790: This string comparison seems fragile (e.g. against "DESC"). Can we use bool or some more restrictive type than this?

ytakaya: I changed the approach to use constants.
`asc_desc_opt` in the syntax file also returns the same constant, so the comparison here is now robust.

k0kubun (approved):
Thanks!

PR #133, opened 2021-07-06
mssqldef: not for replication

ytakaya:
This is a change to support `NOT FOR REPLICATION`.
`NOT FOR REPLICATION` can be used to declare identity columns, check constraints, and foreign key constraints.

The meaning of this syntax can be found in the following document.
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql?view=sql-server-2017
> If this clause is specified for a constraint, the constraint is not enforced when replication agents perform insert, update, or delete operations.


k0kubun (approved):
Thanks!

Issue #11, opened 2018-09-03
Support TRIGGER


k0kubun:
## Problem
We can't specify TRIGGER in input DDLs.

### Initial schema
```
CREATE TABLE users (
  id bigint NOT NULL,
  name text
);

CREATE TRIGGER test BEFORE INSERT ON users
FOR EACH ROW SET name = concat(name, ".");
```

### Input DDLs
```
CREATE TABLE users (
  id bigint NOT NULL,
  name text
);

CREATE TRIGGER test BEFORE INSERT ON users
FOR EACH ROW SET name = concat(name, ".");
```

### Output DDLs
```
-- Nothing is modified --
```

touyu:
Are you planning to implement this in the future?

k0kubun:
Not planned, but I can say at least I still want this feature. So I'd implement it when it becomes critical for me, but if you're planning to implement it in a near term, I'd accept your patch.

k0kubun:
Thanks, @ytakaya! https://github.com/k0kubun/sqldef/pull/135

PR #135, opened 2021-07-14
feature: support trigger

ytakaya:
This is the patch for #11
Now the trigger syntax can be used in mysqldef and mssqldef.

The trigger syntax in postgres needs to call a `FUNCTION` or `PROCEDURE` in `CREATE TRIGGER ...`.
Therefore, in order to use trigger syntax in postgres, we need to support `CREATE FUNCTION ...` and `CREATE PROCEDURE ...` first.
This is future work.
https://www.postgresql.org/docs/12/sql-createtrigger.html


k0kubun (approved):
Wow, great! Thank you :clap:

PR #138, opened 2021-08-05
mssqldef: support cursor in trigger

ytakaya:
This allow the following syntax in the trigger.

- `OPEN cursor`
- `FETCH ... cursor ...`
- `CLOSE cursor`
- `DEALLOCATE cursor`

This pull request is based on https://github.com/k0kubun/sqldef/pull/137.


k0kubun:
re-created at https://github.com/k0kubun/sqldef/pull/145 to minimize the diff

k0kubun:
It seems like the design follows what https://github.com/k0kubun/sqldef/pull/137#issuecomment-892297258 does. I think that's what we want here too.

PR #156, opened 2021-09-22
support table hint

ytakaya:
This is a change to use table_hint.
https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver15

In sqldef, table_hint would be used in `CREATE VIEW AS SELECT ...`. 
For example, in SQL Server, if we assign `with(nolock)` as table_hint in the SELECT statement, we can SELECT the table that is being updated.


k0kubun:
In https://github.com/k0kubun/sqldef/pull/137#issuecomment-891673678, you implied you would only need to fix triggers to satisfy your SQL Server use cases. So I accepted reviewing all these instead of letting you fix and maintain them yourself. But this doesn't seem to be a part of it. Did you already cover your SQL Server use cases with mssqldef and was this just a recent new use case, or are you still working on fixing missing SQL Server support?

ytakaya:
I apologize for my behavior, which contradicts my previous comments.
But you're right, I thought I had exhausted support for SQL Server syntax for existing use cases. This change is due to my carelessness in not being aware of the syntax used in the `CREAT VIEW …`.
I think that this will be the last change to the existing syntax.

Perhaps, as I go through the operation, I may want to support a new syntax again.
However, I think that the change will not be as big as it was for TRIGGER.

k0kubun:
> Perhaps, as I go through the operation, I may want to support a new syntax again.
> However, I think that the change will not be as big as it was for TRIGGER.

If that's the case, could you be a collaborator and maintain the SQL Server support by yourself? I use only MySQL and PostgreSQL, and thus the review cost of every single SQL Server-specific syntax change is non-trivial. Because you're the main user of the SQL Server support, it'd be easier for both of us to let you commit such changes directly.

ytakaya:
Yes, I am the one who uses SQL Server, so it looks like I should be the one to maintain it.
Since I’ll try for careful maintenance, is it possible for you to make me a collaborator?

k0kubun:
Thank you :+1:

> Since I’ll try for careful maintenance, is it possible for you to make me a collaborator?

For sure. I already trust your understanding of the codebase and the design. I sent an invitation to you, so please accept it when you have a chance.

What I expect on you as a maintainer for SQL Server support are:

* You're expected to push commits that are mostly about SQL Server to master without asking my reviews.
  * If you want to leave any description, please write that down in commit messages.
* You're also expected to cut a release that is mostly about SQL Server.
  * If you merge a PR that impacts any behavior, you're expected to cut a release at least once in the same day.
  * It's as easy as updating CHANGELOG.md and running `git tag v0.x.y && git push origin --tags && git push origin master` at the latest master.
  * About versioning, please bump a minor version if you think it's a release with significant or breaking changes, or a teeny version if it's backward-compatible and insignificant. For now, please keep the major version (v0) as is.
* If any issue that is specific to SQL Server or caused by your change is filed, I would like you to fix and close it.

As your first task as a maintainer, I'd like you to resolve the conflict, merge this PR, and make a release.

ytakaya:
Thank you for your detailed explanation, and finished my first task!

PR #159, opened 2021-09-26
Support column comment

JungWinter:
I want to add column with it's comment, so I suggest to add `COMMENT '...'` after `ON UPDATE` keyword.

I usually use only mysql, so if additional patch is needed, let me know anytime. :)

p.s. Thank you for maintaining this tool. I am using it to maintain our db schema as code like IaC.


k0kubun (approved):
Thanks!

Issue #162, opened 2021-09-29
Support for `NEW` keyword in expression


touyu:
## --export output
```sql
CREATE TABLE `users` (
  `id` int unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(255) NOT NULL,
  `deleted_at` timestamp NULL DEFAULT NULL,
  `logical_uniqueness` tinyint(1) DEFAULT '1',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
```

## Input SQL
```sql
CREATE TRIGGER set_logical_uniqueness_on_users BEFORE UPDATE ON users FOR EACH ROW
SET NEW.logical_uniqueness = CASE 
    WHEN NEW.deleted_at IS NULL THEN 1 	
    WHEN NEW.deleted_at IS NOT NULL THEN NULL 
END;
```

## Current output
```
found syntax error when parsing DDL "CREATE TRIGGER set_logical_uniqueness_on_users BEFORE UPDATE ON users FOR EACH ROW
SET NEW.logical_uniqueness = CASE
	WHEN NEW.deleted_at is null then 1
	WHEN NEW.deleted_at is not null then null
END": syntax error at position 128 near 'new'
```

## Expected output
```
-- Apply --
CREATE TRIGGER set_logical_uniqueness_on_users BEFORE UPDATE ON users FOR EACH ROW SET NEW.logical_uniqueness = CASE WHEN NEW.deleted_at IS NULL THEN 1 WHEN NEW.deleted_at IS NOT NULL THEN NULL END;
```

PR #163, opened 2021-09-30
Support for `NEW` keyword in expression

touyu:
Fix: https://github.com/k0kubun/sqldef/issues/162

I considered three implementation methods.

In the end, I adopted the third.

## First Method
Add the following to `colmun_name` in `parser.y`
```
NEW '.' reserved_sql_id
{
  $$ = &ColName{Name: NewColIdent("NEW." + $3.val)}
}
```

This method doesn’t work because the entire `NEW.<colmun_name>` is surrounded with backquotes.

## Second Method
Add the following to `colmun_name` in `parser.y`
```
NEW '.' reserved_sql_id
{
  $$ = &ColName{Qualifier: TableName{Name: NewTableIdent("NEW")}, Name: $3}
}
```

With this method, only`NEW` is surrounded with backquotes
It works, but `NEW` is a keyword, it feels strange to surround it with backquotes.

## Third Method

This pull request.
Add new struct `NewQualifierColName` to avoid surrounding it with backquotes


k0kubun (approved):
Thank you :+1: Great job!

PR #175, opened 2021-12-12
enable to get definition of view with same name, other database for mysql

okaponta:
## What
I added an extra where clause to specify view definition for mysqldef

## Why
 - We have view with the same name with different database.
 - Thus, I added the where clause to specify the view.


k0kubun (approved):
Thanks!

PR #185, opened 2022-01-21
Prefer referencing consts over using string literals

yuku:
This patch was written in favor of boyscout rule.

related comment: https://github.com/yuku/sqldef/commit/833f289f95d63005bcf87ca394ae7007b2f963e8#commitcomment-64216606


k0kubun (approved):
Thanks!

PR #190, opened 2022-01-23
Fix bug that psqldef attempts to drop&add pk every time if two tables have same name and pk

yuku:
Suppose there are two tables having:

* same table name
* same primary key column name

```sql
-- e.g.
create table user.accounts (id int primary key);
create table bank.accounts (id int primary key);
```

psqldef generates following DDL every time:

```sql
-- Apply --
ALTER TABLE "user"."accounts" DROP CONSTRAINT "accounts_pkey";
ALTER TABLE "user"."accounts" ADD primary key ("id");
ALTER TABLE "bank"."accounts" DROP CONSTRAINT "accounts_pkey";
ALTER TABLE "bank"."accounts" ADD primary key ("id");
```

This PR fixes this issue.

### DIFF
--- a/adapter/postgres/postgres.go
+++ b/adapter/postgres/postgres.go
@@ -389,7 +389,7 @@ func (d *PostgresDatabase) getPrimaryKeyColumns(table string) ([]string, error) 
 FROM
 	information_schema.table_constraints AS tc
 	JOIN information_schema.key_column_usage AS kcu
-		ON tc.constraint_name = kcu.constraint_name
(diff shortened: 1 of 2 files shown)


yuku:

FYI: I bet there are almost same issues around `getForeignDefs` but I couldn't write regression tests, so I left it as is.

https://github.com/k0kubun/sqldef/blob/e0b45dd6f40f69d00c119c51c6514e24f8239a11/adapter/postgres/postgres.go#L415-L431

k0kubun (approved):
Thanks!

PR #191, opened 2022-01-24
Make it easy to run a single test case

yuku:
By this change, we can run a test case easily from editor (VSCode in my case).

In addition, the ability to inject envvar has been added, increasing the flexibility of testing.

### Before
image

### After
image


yuku:

k0kubun (approved):
👍

PR #193, opened 2022-01-24
Enable to drop a column even if there is similar column in other schema

yuku:
Suppose there are two similar tables in two schemas:

```sql
create table user.accounts (id int);
create table bank.accounts (id int);
```

If we drop a column from a table:

```sql
create table user.accounts (id int);
create table bank.accounts ();
```

`psqldef` generates duplicated `drop column`:

```
-- Apply --
ALTER TABLE "bank"."accounts" DROP COLUMN "id";
ALTER TABLE "bank"."accounts" DROP COLUMN "id";
```

This PR fixes the issue.


yuku:

k0kubun (approved):
Thanks :pray:

PR #195, opened 2022-01-28
Add `--before-apply` to psqldef

### LINKED ISSUE #192: feature request: enable to prepend `SET ROLE`
It is great if there is a way to insert `SET ROLE` before DDLs. For example, suppose `psqldef` has `--set-role` option:

```
$ psqldef -U github-actions@$PROJECT_ID.iam --set-role owner_role -f schema.sql db_name
-- Apply --
SET ROLE "owner_role";
CREATE TABLE ...
```

### motivation

In postgres, when a resource is created, it is owned by the user/role that executed the associated `CREATE` DDL. Since the owner has special rights to the resource, it is often necessary to control the ownership. If a `psqldef` user wants to set the `owner_role` as the owner of a newly created resource, he can do so by connecting to the postgres server as `owner_role`.

Unfortunately, I can't use `owner_role` directly because:

* My postgres server is running in Google Cloud SQL, and I'm migrating the database continuously using `psqldef` from GitHub Actions.
* To reduce the number of secrets that I need to manage, I want to use [OIDC](https://docs.github.com/en/enterprise-cloud@latest/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform) and [IAM authentication](https://cloud.google.com/sql/docs/postgres/authentication) to connect to Cloud SQL.
* This means that I have to connect to the postgres server as a service account (e.g. `github-actions@PROJECT_ID.iam` ≠ `owner_role`)

### note

Since I'm relatively new to postgres, I may misunderstand the ownership of resources.

### related documents

* [PostgreSQL: Documentation: 14: SET ROLE](https://www.postgresql.org/docs/14/sql-set-role.html)

yuku:
Closes #192 

This PR adds `--before-apply` option to psqldef as discussed in #192.

### Screenshots
image

image

### DIFF
--- a/sqldef.go
+++ b/sqldef.go
@@ -17,6 +17,7 @@ type Options struct {
 	DryRun      bool
 	Export      bool
 	SkipDrop    bool
+	BeforeApply string
 }
 
 // Main function shared by `mysqldef` and `psqldef`
@@ -51,12 +52,20 @@ func Run(generatorMode schema.GeneratorMode, db adapter.Database, options *Optio
 		return
 	}
 
+	beforeApply := []string{}
+	for _, query := range strings.Split(options.BeforeApply, ";") {
+		query = strings.TrimSpace(query)
+		if len(query) > 0 {
+			beforeApply = append(beforeApply, query)
+		}
+	}
+
 	if options.DryRun || len(options.CurrentFile) > 0 {
-		showDDLs(ddls, options.SkipDrop)
+		showDDLs(ddls, options.SkipDrop, beforeApply)
 		return
 	}
 
-	err = adapter.RunDDLs(db, ddls, options.SkipDrop)
+	err = adapter.RunDDLs(db, ddls, options.SkipDrop, beforeApply)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


yuku:
@k0kubun This PR adds `--before-apply` to psqldef only. Should I add the option to other commands too?

sqldef.go
```diff
@@ -51,12 +52,20 @@ func Run(generatorMode schema.GeneratorMode, db adapter.Database, options *Optio
 		return
 	}
 
+	beforeApply := []string{}
+	for _, query := range strings.Split(options.BeforeApply, ";") {
```
k0kubun, line 56: This could wrongly split an SQL like `... xxx = ";";`. Can we just NOT split this?

k0kubun:
> This PR adds --before-apply to psqldef only. Should I add the option to other commands too?

No, it's fine. Until somebody comes up with a use case, we shouldn't have one.

sqldef.go
yuku: addressed in 58561f7

yuku:
It seems the failed test case isn't related to this PR.

k0kubun (approved):
Thank you!

PR #196, opened 2022-01-29
Enable to drop and add check constraints correctly

yuku:
`psqldef` couldn't handle adding and dropping check constraint to a table. This PR fixes the issue.

Suppose there is a following table in Postgres:

```sql
CREATE TABLE foo (
  bar INT CHECK (bar > 0)
);
```

In master, `adapter.postgres.PostgresDatabase.DumpTableDDL()` returns a string like below for the table:

```sql
CREATE TABLE foo (
  bar INT,
  CHECK (bar > 0)
);
```

so the sqlparser think the `CHECK (bar > 0)` belongs to the table not to the `bar` column. Thus sqldef couldn't handle check constraints correctly.

To address the issue, this PR fixes the `DumpTableDDL()` to return the following string:

```sql
CREATE TABLE foo (
  bar INT CONSTRAINT foo_bar_check CHECK (bar > 0)
);
```

This PR also fixes schema/generator.go to handle column checks correctly.


yuku:
@k0kubun could you review this PR? 🙏

k0kubun:
I'm going to work on CI failures and PRs from now.

Meanwhile, I'd like to ask a question. How many other things in psqldef do you plan to fix? While I appreciate the fact that you create well-written pull requests instead of issues and fix a lot of things, the cost of reviewing and merging pull requests is not zero. This is what I've told to people who raised multiple issues / pull requests for multiple days: I'd like to see all issues or pull requests you have at once instead of seeing one every day because I have other OSS and tasks to be taken care of and I'd like to predict the effort taken to satisfy each contributor.

yuku:
Sorry for bothering you. The reason why I had split the changes into series of pull requests was in the hope that it would make the review process less burdensome. Next time I will try to send patches in batches if possible.

> How many other things in psqldef do you plan to fix?

At this moment, I recognize that there are following issues around psqldef for my use case:

* Syntax support
  * (high priority): [Indexes on Expressions](https://www.postgresql.org/docs/14/indexes-expressional.html), especially function index.
    * back ground: my database has a table that stores JSON returned by Slack (known as [`Installation`](https://github.com/slackapi/node-slack-sdk/blob/4989e7f827640afd5f5d25e40750d7030ca9092a/packages/oauth/src/index.ts#L645-L709)) and I want to add a unique constraint on `slack_app_installation->'team'->>'id'` or `jsonb_extract_path_text(slack_app_installation, 'team', 'id')`
    * why hight?: I can extract the team id in application side and store it in another column and add unique constraint to it. But I don't want to do such denormalization.
  * (middle priority): array support in check condition.
    * back ground: I have a table that has `state` column and some other columns values should change depending on the `state`. To test the rule, I'd like to write `CHECK ((state in ('a', 'b') and user_id is null) or (state not in ('a, 'b') and user_id is not null))`.
      * note: psqldef can accept the definition, but once it is store in postgres, postgres converts it as `CHECK (state = ANY(ARRAY['a', 'b']) AND user_id IS NULL ...)`. psqldef cannot parse `ARRAY` literal now.
    * why middle?: I can rewrite the definition as `CHECK ((state = 'a' or state = 'b') and user_id is null) or (state != 'a' and state != 'b') and user_id is not null)`. As long as the number of state is small this work around works.
  * (low priority): `DEFAULT function_call()`.
    * back ground: I'm using UUID for id column so I'd like to define as `id UUID DEFAULT uuid_generate_v4()`.
    * why low?: there is a work around `insert into foo (id) values (uuid_generate_v4())`
* Bug fixes
  * (high priority): `alter table non_public.foo add constraint uniq unique (col);` always generate `DROP INDEX uniq;` and it causes `pg: index "uniq" does not exist`.
    * why high?: the table storing slack installation JSON lives in non public schema.

By the way, it is hard for me to estimate the number of bug fixes I need in the future.

yuku:
How about stating your policy in README.md or CONTRIBUTING.md? I think it is the way around GitHub ecosystem.

k0kubun:
> Next time I will try to send patches in batches if possible.

Thank you! Note that it doesn't mean I'm expecting you to include all changes in a single PR. I still want you to file pull requests with the same granularity. Since it's impossible to make chained pull requests unless you own the repository, you could do that by including change1 in PR1, change1 and 2 in PR2, and change1, 2, and 3 in PR3, and so on. Once I merge PR1 to master and PR2, I can see only the diff of change2 in PR2. So "it would make the review process less burdensome" too.

> By the way, it is hard for me to estimate the number of bug fixes I need in the future.

I understand. I'm assuming you could at least notice bugs that happen with your current schema while writing patches, and you will not encounter bugs until you newly use a DDL feature you haven't used or somebody else files a patch that has a bug. So the frequency would be much lower. That is enough for me.

> How about stating your policy in README.md or CONTRIBUTING.md? I think it is the way around GitHub ecosystem.

done https://github.com/k0kubun/sqldef#contributing. Note that I use GitHub as you do at work, but nowadays the time I can use for OSS maintenance is pretty limited because I'm a full-time worker, a graduate student (in the last semester, fortunately), and a father of a 1-year old kid at the same time, while I also relatively quickly respond to every GitHub notification that I constantly get everywhere. So, for making my OSS maintenance sustainable, having a large chunk of work at once is so helpful for me to reduce context switches. Thank you for your patience in advance.

k0kubun (approved):
I'm looking forward to seeing your remaining contributions. Thanks!

PR #197, opened 2022-01-29
Enable to add a unique constraint to tables in non-public schema

yuku:
Fixes the bug I mentioned in https://github.com/k0kubun/sqldef/pull/196#issuecomment-1024840782

> * Bug fixes
>   * (high priority): `alter table non_public.foo add constraint uniq unique (col);` always generate `DROP INDEX uniq;` and it causes `pg: index "uniq" does not exist.`
>     * why high?: the table storing slack installation JSON lives in non public schema.

This PR also includes two chore commits:
1. c693d02031234f590f6ccd26b04f261f0ac7e805 While addressing the issue, I needed to check the result of `--export` many times, so I added `assertExportOutput()` helper and refactored psqldef_test.go a bit.
2. 31542f5661b80b6585bebdde266cb5109ab90d72 When I tested the `assertExportOutput()`, I found a bug in the test case added by #196
    * It was a test of a situation that rarely happens, so I simply dropped it.

NOTE: As I explained in https://github.com/k0kubun/sqldef/pull/196#issuecomment-1024840782, I'm planning to send one more pull request for the other high priority issue for me in the near future.

### DIFF
--- a/adapter/postgres/postgres.go
+++ b/adapter/postgres/postgres.go
@@ -152,10 +152,14 @@ func (d *PostgresDatabase) DumpTableDDL(table string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return buildDumpTableDDL(table, cols, pkeyCols, indexDefs, foreignDefs, policyDefs, checkConstraints), nil
+	uniqueConstraints, err := d.getUniqueConstraints(table)
+	if err != nil {
+		return "", err
+	}
+	return buildDumpTableDDL(table, cols, pkeyCols, indexDefs, foreignDefs, policyDefs, checkConstraints, uniqueConstraints), nil
 }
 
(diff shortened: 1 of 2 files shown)


adapter/postgres/postgres.go
```diff
@@ -345,19 +352,30 @@ func (d *PostgresDatabase) getColumns(table string) ([]column, error) {
...
+	AND    indexName NOT IN (SELECT name FROM unique_and_pk_constraints)
+	`
 	schema, table := SplitTableName(table)
 	rows, err := d.db.Query(query, schema, table)
 	if err != nil {
 		return nil, err
 	}
 	defer rows.Close()
 
-	uniqueConstraints, err := d.getUniqueConstraints(table)
```
yuku, line 356: Since `rows.Close()` is deferred and has not been executed yet at this line, `d.getUniqueConstraints(table)` always returns an empty slice unexpectedly.

k0kubun (approved):
Thanks!

PR #199, opened 2022-01-31
Support creating indexes on expressions and using function as default

yuku:
* 1668c76458dfb1a2823713eb3dbb1872189a5316 Support creating indexes on expressions:

    ```sql
    create table test (col jsonb);
    create unique index uniq on test (jsonb_extract_path_text(col, 'foo', 'bar'));
    --                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ```
    * The high priority thing described in [this comment](https://github.com/k0kubun/sqldef/pull/196#issuecomment-1024840782).
    * If `func_name(col_name, 'param1')` is given as index, Postgres internally stores it as `func_name(col_name, VARIADIC ARRAY['param1'::text])`. So this PR also adds simple [array constructor](https://www.postgresql.org/docs/14/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS) syntax support to psqldef. 964d9956a4e7ddd5ba65906594d1d8fc52428dad
* 8f21a053b95c8811f3149e523e3d044d6a24b238 Support using function as default:

    ```sql
    create table test (id uuid default generate_uuid_v4());
    --                                 ^^^^^^^^^^^^^^^^^^
    ```

    * The low priority thing described in [this comment](https://github.com/k0kubun/sqldef/pull/196#issuecomment-1024840782).
* 2fd5f9db70cf3193eef4445b302a3add89621e27 Enable to drop index form tables in non-public table.
    * I found the bug while writing a test case for other changes.

**Chore**
* Added `publicAndNonPublicSchemaTestCases` variable to postgres_test.go
  * It makes us easy to write same test case for both public and non-public schemas.


yuku:
Let me add a bug fix commit 🙇 

In master, `create view non_public.foo ...` always cause `create or replace view public.foo ...`. 38b6f1c fixes the issue.

k0kubun (approved):
Thanks!

Issue #212, opened 2022-03-02
[mssqldef] index exported without ';', identifier exported without '[' ']'


icedac:
Hi I'm back with two more :)
 
 
## sample mssql db
```
SET ANSI_NULLS ON
GO
SET QUOTED_IDENTIFIER ON
GO
CREATE TABLE [dbo].[Table_1](
	[no] [nvarchar](50) NULL,
	[type] [bigint] NULL,
	[create] [int] NULL
) ON [PRIMARY]
GO

SET ANSI_NULLS ON
GO
SET QUOTED_IDENTIFIER ON
GO
CREATE TABLE [dbo].[Table_2](
	[ttt] [nchar](10) NULL
) ON [PRIMARY]
GO
SET ANSI_PADDING ON
GO

CREATE NONCLUSTERED INDEX [IX_Table_1] ON [dbo].[Table_1]
(
	[no] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
GO
CREATE NONCLUSTERED INDEX [IX_Table_1_1] ON [dbo].[Table_1]
(
	[type] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, DROP_EXISTING = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
GO
```

## Current export output
```
CREATE TABLE dbo.Table_1 (
    no nvarchar(50),
    type bigint,
    create int
);
[...]

PR #213, opened 2022-03-04
[mssqldef] fix for export index, reserved keyword named table column

icedac:
#212 crappy fix working for me
 
 
@ytakaya check this ugly (2 days golang coder) fix 
  
  
It figured out 'type', 'no' column name have problem with sql parser not sqlserver


adapter/mssql/mssql.go
```diff
@@ -14,6 +14,725 @@ import (
 
 const indent = "    "
 
+// names from: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql?view=sql-server-ver15
+var reservedKeywordsArray = [...]string{
```
k0kubun, line 18: Isn't just escaping every identifier sufficient? If not, could we possibly reuse this table? https://github.com/k0kubun/sqldef/blob/888f344ebf388d5224eb3eb69a9299e4a751ae17/sqlparser/token.go#L94

icedac: I think escaping everything is better. I didn't because I read only small parts of source code so was afraid breaking something.

icedac: And now I knew that token keywords map are root of my original issue.

icedac:
I add one more
 
easy Export and then Import for mssqldef
- sort table order by fk dependency prevents fresh new import from exported one

ytakaya:
@icedac 
I think this PR is packed with solutions to many issues.
I found
- Issue when using reserved word column names (Is this the problem you wanted to solve?).
- Issue on view output.
- Issue on the dependence of fk.

Can you separate the PRs per issue so that the changes are easy to understand?

k0kubun:
Given that it has a conflict now and it's waiting for a revision, let me close this until this becomes actionable on our end. Please reopen this when it's ready. Thank you!

PR #218, opened 2022-03-11
Add features to avoid environmental dependence

tadamatu:
I have added 3 features to avoid environment dependence.

# --targets, --target-file  (For All)
```
--targets                     Manage the target name (Table, View, Type, Trigger)
--target-file                 File management of --targets option
```
This feature can be manage target tables.
Managing with files makes it easier to manage tables.
For example, if there are tables such as `users_backup` or `items_20220301` only in production, you can ignore them.
See `README.md` for a sample.

Review Tips => commit: b627290807dfdeb4bb7ce84881860103eb57a736

# --without-partition-range (For MySQL)
```
--without-partition-range     Without the specific code of PARTITION BY RANGE
```
In MySQL, partition information exists in DumpDDL.
Currently, its presence causes a syntax error, which can be avoided by this feature.
For example, if manage partitions by date and there are differences in the environment, it can be ignore them.
See `README.md` for a sample.

Review Tips => commit: 1fe7e3f6d5096d05fd62b57165a016a248e3aebd

# --init-auto-increment   (For MySQL)
```
--init-auto-increment         Initialize AUTO_INCREMENT for CREATE TABLE
```
MySQL has an `AUTO_INCREMENT` count in DumpDDL.
We want to clear this count when we execute `CREATE TABLE` across environments. This feature does that.
See `README.md` for a sample.

Review Tips => commit: e9b5868501b8ffe940a27810527cd9932e7a8482

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Application Options:
       --export                      Just dump the current schema to stdout
       --skip-drop                   Skip destructive changes such as DROP
       --without-partition-range     Without the specific code of PARTITION BY RANGE
+      --init-auto-increment         Initialize AUTO_INCREMENT for CREATE TABLE
       --help                        Show this help
       --version                     Show this version
 ```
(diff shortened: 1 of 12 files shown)


tadamatu:
CI errors are due to different schemas depending on the version of MySQL.
`MySQL8.0` becomes` bigint`
However, `MySQL5.7` becomes` bigint (20) `.
I'll think a little, so please wait.
Or it would be helpful if you could give me some good ideas.

k0kubun:
> CI errors are due to different schemas depending on the version of MySQL

When the test is not mainly about the new CLI flags but bignum, please use `cmd/*def/tests.yml` to test it. You can use `min_version: '8.0'` or `max_version: '5.7'` to control target versions.

k0kubun:
> I have added 3 features to avoid environment dependence.

Could you file 3 pull requests instead of having them all in one? It's hard to review your pull requests currently. To avoid future conflicts, you may file it like:

* pull request 1: commit1
* pull request 2: commit1 commit2
* pull request 3: commit1 commit2 commit3

Then, if I merge it like pull request 1, pull request 2, and then pull request 3, I will be able to look at the diff of only one feature at a time.

tadamatu:
@k0kubun 

> When the test is not mainly about the new CLI flags but bignum, please use cmd/*def/tests.yml to test it. You can use min_version: '8.0' or max_version: '5.7' to control target versions.

thanks. I'll try it.

> Could you file 3 pull requests instead of having them all in one? 

I see. This PR will be closed to split.

PR #221, opened 2022-03-14
feat: support for managing target table names

tadamatu:

# --target-file  (For All)
```
--target-file                Manage the target name (Table, View, Type, Trigger)
```
This feature can be manage target tables.
Managing with files makes it easier to manage tables.
For example, if there are tables such as `users_backup` or `items_20220301` only in production, you can ignore them.
See `README.md` for a sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Application Options:
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
       --skip-drop                   Skip destructive changes such as DROP
+      --targets                     Manage the target name (Table, View, Type, Trigger)
+      --target-file                 File management of --targets option
       --help                        Show this help
       --version                     Show this version
 ```
@@ -98,6 +100,33 @@ $ mysqldef -uroot test --skip-drop < schema.sql
 Skipped: 'DROP TABLE users;'
 ```
 
+#### Options to avoid environment dependence
+
+These options may be useful when running in different environments (e.g. stg <->prod)
+
+##### --targets, --target-file
+
+Tables other than users, orders, mails are ignored
+
+```sql
+$ mysqldef -uroot test --targets users,orders,mails --export > schema.sql
+$ mysqldef -uroot test --targets users,orders,mails < schema.sql
+```
+
+The following works the same as the command above
+
+```plaintext
+# Save as file with the name `tables_file`
+users
+orders
+mails
+```
+
+```sql
+$ mysqldef -uroot test --target-file tables_file --export > schema.sql
+$ mysqldef -uroot test --target-file tables_file < schema.sql
+```
+
 ### psqldef
 
 `psqldef` should work in the same way as `psql` for setting connection information.
@@ -117,6 +146,8 @@ Application Options:
       --dry-run              Don't run DDLs but just show them
       --export               Just dump the current schema to stdout
       --skip-drop            Skip destructive changes such as DROP
+      --targets              Manage the target name (Table, View, Type, Trigger)
+      --target-file          File management of --targets option
       --help                 Show this help
 ```
 
@@ -197,6 +228,8 @@ Application Options:
       --dry-run          Don't run DDLs but just show them
       --export           Just dump the current schema to stdout
       --skip-drop        Skip destructive changes such as DROP
+      --targets          Manage the target name (Table, View, Type, Trigger)
+      --target-file      File management of --targets option
       --help             Show this help
 ```
 
@@ -216,6 +249,8 @@ Application Options:
       --dry-run              Don't run DDLs but just show them
       --export               Just dump the current schema to stdout
       --skip-drop            Skip destructive changes such as DROP
+      --targets              Manage the target name (Table, View, Type, Trigger)
+      --target-file          File management of --targets option
       --help                 Show this help
       --version              Show this version
 ```
--- a/README.md
+++ b/README.md
@@ -36,8 +36,7 @@ Application Options:
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
       --skip-drop                   Skip destructive changes such as DROP
-      --targets                     Manage the target name (Table, View, Type, Trigger)
-      --target-file                 File management of --targets option
+      --target-file                 Manage the target name (Table, View, Type, Trigger)
       --help                        Show this help
       --version                     Show this version
 ```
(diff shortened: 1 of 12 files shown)


k0kubun:
Sorry for causing trouble, but could you also file two other files as well at once? I'd like to see all pull requests at once while I also prefer seeing them as separated pull requests.

tadamatu:
> Sorry for causing trouble, but could you also file two other files as well at once? I'd like to see all pull requests at once while I also prefer seeing them as separated pull requests.

I made 3 pull request.
Are those the same as your want?

k0kubun:
Yes, that's what I exactly expected. Thank you for filing pull requests this way :)

k0kubun:
General feedback to all the features: sqldef tries to imitate the original `*sql` CLI without `def` whenever possible because using the same interface makes it easier for users to understand how to use it. So I'm generally reluctant to introduce CLI options that don't exist in the original CLI. When you need to fix some problem in sqldef, it's preferable to fix the problem without introducing sqldef's original options. 

Cloud you think about the solution again for all of the pull requests, keeping the above idea in mind? For example in this PR, do you really need both `--targets` and `--target-file`? Introducing only one is better than two, introducing only options that exist in the original CLI is better than original one, and having zero option is the most ideal solution.

k0kubun:
> BIN +7.57 MB [cmd/mysqldef/mysqldef_linux_amd64](https://github.com/k0kubun/sqldef/pull/221/files#diff-c993fb5d64e7e05cc7de2aea845ab01df4361cd3bb57614a41c5e4f28c5524e1)

also, please avoid committing a binary.

tadamatu:
> also, please avoid committing a binary.

very sorry...I removed it.

> Cloud you think about the solution again for all of the pull requests, keeping the above idea in mind? For example in this PR, do you really need both --targets and --target-file? Introducing only one is better than two, introducing only options that > exist in the original CLI is better than original one, and having zero option is the most ideal solution.

I think so.
But when it comes to this PR feature, I think it useful.
Is it possible to adopt either one?

tadamatu:
I'll write more clearly why I implemented this.

Initially I developed only `--targets`.
I wanted to narrow down target tables.

I believe that garbage such as `users_bk` ` users_20220301` is likely to exist locally.
It will be especially effective when multiple people develop using the same database.

Also, I feel that `sqldef` has potential and it is regrettable to use it experimentally only locally.
I believe it should be useful to narrow down targets so that no one makes a mistake. Like `.gitignore` (although the approach is the opposite)
I personally think it's a relatively used feature when using `sqldef` in a project. 

And when using it in a project, I wanted to manage targets as a file, so I developed `--target-file`.

k0kubun:
I'm not questioning the usefulness of this feature. I think this is an important feature to support. I just need to be careful introducing any original option no matter what it is. And I need more time to think about it since you're proposing two, not just one of such options. The same applies to https://github.com/k0kubun/sqldef/pull/222; as long as you're proposing to fix a solution by introducing yet another original option, I must think hard about how we could fix it without it.

Meanwhile, if you think these are important problems to be solved, please keep thinking about minimizing the number of new options https://github.com/k0kubun/sqldef/pull/221#issuecomment-1067534374. The fewer options a pull request has, the more quickly it will be merged. I read https://github.com/k0kubun/sqldef/pull/221#issuecomment-1071293128 but it still doesn't really justify why you didn't drop `--targets` when you added `--target-file`.

tadamatu:
> I'm not questioning the usefulness of this feature. I think this is an important feature to support.

Thank you for understanding!

I agree with reducing the options.
I hadn't removed the option yet because I'm wondering which one I should remove. This is because I wanted to have a conversation with you before working on it.
In my opinion, `--target-file` option was better in terms of ease of use for a project, so I left that one. 
(removed `--targets` option)

6ca55fd265d5d81ce67bc86ad10abe445e0dd889

k0kubun:
I'm sorry for taking such a long time to respond to this pull request. I'll be still busy for the upcoming four weeks, but I'm moving this thread forward since I've already let you wait too long.

---

I agree that it's a legitimate use case and this is something not fixable with an input SQL. So we likely need to introduce some extra config here. However, I need you to make some changes to this pull request before merging it:

* Can you add the option to only the sqldef commands you actually use?
  * For example, if you use only MySQL, please add the option to only `mysqldef`.
  * While I value consistency across commands, I'd like to be careful about the interface that doesn't exist in the original CLI.
  * Also, I'm still not 100% confident about using this design to support it, so I'd like to minimize the potential backward incompatibility we might need to make in the future.
     * In particular, I currently consider introducing something like `--config sqldef.yml` that allows you to specify stuff in the file in addition to CLI options. In that config, I can be less concerned about compatibility with the original CLI.
* Can we invert how we specify the targets? In other words, instead of `--target-file`, can we introduce `--skip-file` that lists targets to be ignored?
  * Thinking about the use case, I think it'd be easier to ignore `users_backup` or `items_20220301` with this.
  * To support ignoring tables like `items_20220301`, I think the input should support regex. 
     * For the convenience of use without regex, each line should probably get `^` and `$` at the beginning and the end respectively so that you don't need to write it for every line.
* Can we limit the control scope to tables?
  * This is for simplicity and the backward compatibility problem that I mentioned above. I think the use case is to ignore backup tables, so it seems enough to only support tables.

Please feel free to ask questions if you have opinions about my suggestions. Otherwise, please leave a comment after you fix those three things in your patch.

tadamatu:
@k0kubun 
I will close this pull request because I created another pull request.
https://github.com/k0kubun/sqldef/pull/242

> In particular, I currently consider introducing something like --config sqldef.yml that allows you to specify stuff in the file in addition to CLI options. In that config, I can be less concerned about compatibility with the original CLI.

I like this idea.
The feature I hope is that filter tables.
If config includes that feature, it's okay to close the above pull request.

thank you.

k0kubun:
> Can we invert how we specify the targets? In other words, instead of --target-file, can we introduce --skip-file that specifies a file that lists targets to be ignored?

Based on the input at https://github.com/k0kubun/sqldef/issues/246#issuecomment-1153115287, I ended up implementing something closer to your original proposal, `--target-file`, at https://github.com/k0kubun/sqldef/pull/250.

@tadamatu Once it's released, which feature would you use, `--skip-file` or `target_tables`? I'm going to remove `--skip-file` from the options in favor of `--config`, and if you prefer the `--skip-file` feature, I can support `skip_tables` in `--config` in addition to `target_tables`. If you'd rather prefer `target_tables`, I'd like to only leave `target_tables`.

k0kubun:
As I already implied at https://github.com/k0kubun/sqldef/pull/221#issuecomment-1097571408 the removal of that option when we introduce a config YAML and I've got no response for 10 days, I'm going to remove the option and leave only `--config` https://github.com/k0kubun/sqldef/pull/250 in the upcoming release. Please let me know if you want `skip_tables` in the config interface, but potentially you could do the same thing in a regexp in `target_tables`.

k0kubun:
`skip_tables` has been added back to `--config` in v0.13.19.

PR #222, opened 2022-03-14
feat: [mysqldef] avoid environment-dependent for PARTITION BY RANGE

tadamatu:

# --ignore-partition-range (For MySQL)
```
--ignore-partition-range     Ignore the specific code of PARTITION BY RANGE
```
In MySQL, partition information exists in DumpDDL.
Currently, its presence causes a syntax error, which can be avoided by this feature.
For example, if manage partitions by date and there are differences in the environment, it can be ignore them.
See `README.md` for a sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Application Options:
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
       --skip-drop                   Skip destructive changes such as DROP
+      --targets                     Manage the target name (Table, View, Type, Trigger)
+      --target-file                 File management of --targets option
       --help                        Show this help
       --version                     Show this version
 ```
@@ -98,6 +100,33 @@ $ mysqldef -uroot test --skip-drop < schema.sql
 Skipped: 'DROP TABLE users;'
 ```
 
+#### Options to avoid environment dependence
+
+These options may be useful when running in different environments (e.g. stg <->prod)
+
+##### --targets, --target-file
+
+Tables other than users, orders, mails are ignored
+
+```sql
+$ mysqldef -uroot test --targets users,orders,mails --export > schema.sql
+$ mysqldef -uroot test --targets users,orders,mails < schema.sql
+```
+
+The following works the same as the command above
+
+```plaintext
+# Save as file with the name `tables_file`
+users
+orders
+mails
+```
+
+```sql
+$ mysqldef -uroot test --target-file tables_file --export > schema.sql
+$ mysqldef -uroot test --target-file tables_file < schema.sql
+```
+
 ### psqldef
 
 `psqldef` should work in the same way as `psql` for setting connection information.
@@ -117,6 +146,8 @@ Application Options:
       --dry-run              Don't run DDLs but just show them
       --export               Just dump the current schema to stdout
       --skip-drop            Skip destructive changes such as DROP
+      --targets              Manage the target name (Table, View, Type, Trigger)
+      --target-file          File management of --targets option
       --help                 Show this help
 ```
 
@@ -197,6 +228,8 @@ Application Options:
       --dry-run          Don't run DDLs but just show them
       --export           Just dump the current schema to stdout
       --skip-drop        Skip destructive changes such as DROP
+      --targets          Manage the target name (Table, View, Type, Trigger)
+      --target-file      File management of --targets option
       --help             Show this help
 ```
 
@@ -216,6 +249,8 @@ Application Options:
       --dry-run              Don't run DDLs but just show them
       --export               Just dump the current schema to stdout
       --skip-drop            Skip destructive changes such as DROP
+      --targets              Manage the target name (Table, View, Type, Trigger)
+      --target-file          File management of --targets option
       --help                 Show this help
       --version              Show this version
 ```
--- a/README.md
+++ b/README.md
@@ -142,14 +142,6 @@ CREATE TABLE user (
 ```
 
 ```sql
-$ mysqldef -uroot test --export --ignore-partition-range > schema.sql
-CREATE TABLE user (
-  id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
-  name VARCHAR(128) DEFAULT 'user1',
-) Engine=InnoDB DEFAULT CHARSET=utf8mb4;
-```
-
-```sql
 $ mysqldef -uroot test --export > schema.sql
 $ mysqldef -uroot test --dry-run --ignore-partition-range < schema.sql
 Run: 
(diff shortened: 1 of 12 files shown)


k0kubun:
> Currently, its presence causes a syntax error
> For example, if manage partitions by date and there are differences in the environment, it can be ignore them.

I want to know more about the background. First, suppose a situation where sqldef can correctly parse partition ranges. Would that fix your problem, or would sqldef wrongly detect changes in partition ranges every time when you manage partitions by date?

tadamatu:
> > Currently, its presence causes a syntax error
> > For example, if manage partitions by date and there are differences in the environment, it can be ignore them.
> 
> I want to know more about the background. First, suppose a situation where sqldef can correctly parse partition ranges. Would that fix your problem, or would sqldef wrongly detect changes in partition ranges every time when you manage partitions by date?

If sqldef can parse the partition range correctly, some of my concerns will be resolved.

However, it may not be possible to resolve the following cases.
For example, in prod, partitions are being added programmatically every day. (by batch etc.)
However, in local, it makes a difference because partitions is not added.
Normally, the schema is reflected in local, so there will be differences. (like Local -> dev -> stg -> prod)
Since MySQL has the range of partitions in the schema, I think there may be cases what want to ignore it.

tadamatu:
I received a [PR comment](https://github.com/k0kubun/sqldef/pull/223#issuecomment-1067530640) from `AUTO_INCREMENT=xxx` and decided that it was also unnecessary, so I removed it from the` --export` process.

db77387ab4a58286b2496cd4e36fa953f51853b5

k0kubun:
> I received a https://github.com/k0kubun/sqldef/pull/223#issuecomment-1067530640 from AUTO_INCREMENT=xxx and decided that it was also unnecessary, so I removed it from --export process.

I got confused because I thought this is the second PR and so it shouldn't be included in this PR. However, despite the pull request number (https://github.com/k0kubun/sqldef/pull/221 → https://github.com/k0kubun/sqldef/pull/222 → https://github.com/k0kubun/sqldef/pull/223), I guess it's actually https://github.com/k0kubun/sqldef/pull/221 → https://github.com/k0kubun/sqldef/pull/223 → https://github.com/k0kubun/sqldef/pull/222. It would have been a bit easier to understand it if you filed pull requests in order, but I'm okay with if you can confirm the above understanding is correct.

> However, it may not be possible to resolve the following cases.
> For example, in prod, partitions are being added programmatically every day. (by batch etc.)
> However, in local, it makes a difference because partitions is not added.
> Normally, the schema is reflected in local, so there will be differences. (like Local -> dev -> stg -> prod)
> Since MySQL has the range of partitions in the schema, I think there may be cases what want to ignore it.

I'm not sure if I follow your scenario correctly. Could you show an example of the expected schema for each step in SQL? I'd like to _at least_ see what's different between production and development in SQL.

tadamatu:
> I'm not sure if I follow your scenario correctly. Could you show an example of the expected schema for each step in SQL? I'd like to at least see what's different between production and development in SQL.

For example, suppose there is such schema in `prod` env.
```sql
CREATE TABLE `users` (
  `uuid` varchar(37) NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  `joined` date NOT NULL,
  PRIMARY KEY (`uuid`,`joined`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
/*!50100 PARTITION BY RANGE (year(`joined`))
(PARTITION p202109 VALUES LESS THAN ('2021-10-01'),
 PARTITION p202110 VALUES LESS THAN ('2021-11-01'),
 PARTITION p202111 VALUES LESS THAN ('2021-12-01'),    -- added by batch 
 PARTITION p202112 VALUES LESS THAN ('2022-01-01'),    -- added by batch
 PARTITION pmax VALUES LESS THAN MAXVALUE) */;
```

`local` env at this point.
```sql
CREATE TABLE `users` (
  `uuid` varchar(37) NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  `joined` date NOT NULL,
  PRIMARY KEY (`uuid`,`joined`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
/*!50100 PARTITION BY RANGE (year(`joined`))
(PARTITION p202109 VALUES LESS THAN ('2021-10-01'),
 PARTITION p202110 VALUES LESS THAN ('2021-11-01'),
 PARTITION pmax VALUES LESS THAN MAXVALUE) */;
```

Development has progressed in `local` env.
```sql
CREATE TABLE `users` (
  `uuid` varchar(37) NOT NULL,
  `name` varchar(255) DEFAULT NULL,
  `age` int,                                              -- added
  `joined` date NOT NULL,
  PRIMARY KEY (`uuid`,`joined`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
/*!50100 PARTITION BY RANGE (year(`joined`))
(PARTITION p202109 VALUES LESS THAN ('2021-10-01'),
 PARTITION p202110 VALUES LESS THAN ('2021-11-01'),
 PARTITION pmax VALUES LESS THAN MAXVALUE) */;
```

Then, I want to apply the schema developed in local env to production env. 
Export the local DDL and...
```console
mysqldef -h 127.0.0.1 -uroot -pXX --export > schema.sql
```

Apply to production.
```console
mysqldef -h prod.domain -uroot -pXX < schema.sql
```

If sqldef were to be able to parse the partition ranges, what would happen to the DDL executed at this time?
(1) or (2) ?
I expect (1).

(1)
```sql
ALTER TABLE `users` ADD `age` int AFTER `name`;
```
(2)
```sql
ALTER TABLE `users` ADD `age` int AFTER `name`;
ALTER TABLE `users` DROP PARTITION p202111;
ALTER TABLE `users` DROP PARTITION p202112;
```

tadamatu:
`PARTITION BY RANGE` has many uses: range, hash, interval, composite...
Some are fixed during the initial DB migration, others are added automatically during daily operations.
The two are in conflict.
some may be happy to have `PARTITION BY RANGE` in DDL and others may not.

I think a decision needs to be made as to which should be the default, 
Personally, I think it's better to include `PARTITION BY RANGE`  in the DDL by default, and optionally ignore it.
（Currently, partition range is not supported, so there is an idea to default to ignore without options）

What is your opinion?
I would like to have a little discussion before working on it.

k0kubun:
Sorry, it seems like I get a higher volume of GitHub notifications for OSS than usual (including yours) these days and I don't afford to respond to ones that require effort to decide a design right away. Please wait for me a little, and keep thinking about alternatives that don't require a new option. Thank you.

k0kubun:
Please try v0.11.48 and let me know if it still misses your use case.

tadamatu:
@k0kubun 
Sorry for the delay in noticing. (I've been a little busy personally...)
I will try v0.11.48 !

I understood your policy very well through some PRs.
If I can make time for the challenge, I will try to submit another PR.
Thank you very much!

PR #223, opened 2022-03-15
feat: [mysqldef] avoid environment-dependent for AUTO_INCREMENT

tadamatu:

# --init-auto-increment   (For MySQL)
```
--init-auto-increment         Initialize AUTO_INCREMENT for CREATE TABLE
```
MySQL has an `AUTO_INCREMENT` count in DumpDDL.
We want to clear this count when we execute `CREATE TABLE` across environments. This feature does that.
See `README.md` for a sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Application Options:
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
       --skip-drop                   Skip destructive changes such as DROP
+      --targets                     Manage the target name (Table, View, Type, Trigger)
+      --target-file                 File management of --targets option
       --help                        Show this help
       --version                     Show this version
 ```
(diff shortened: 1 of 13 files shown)


k0kubun:
> We want to clear this count when we execute CREATE TABLE across environments.

I get the motivation, but why don't you just remove `AUTO_INCREMENT=xxx` in your schema file after exporting it? Do you need to always maintain the schema file using `--export`?

tadamatu:
> I get the motivation, but why don't you just remove `AUTO_INCREMENT=xxx` in your schema file after exporting it? Do you need to always maintain the schema file using `--export`?

That's true!
Remove that code from `--export` process.

k0kubun:
It seems like I failed to let you understand what I was thinking about. I'm sorry.

First off, as I wrote in https://github.com/k0kubun/sqldef/pull/221#issuecomment-1067534374, I don't want to introduce CLI options that don't exist in `mysql` to `mysqldef` unless it's 100% impossible to solve a problem without it. And out of the options you've proposed, I think `--init-auto-increment` is the easiest to remove while still solving your problem. So I'll likely merge this PR once `--init-auto-increment` is removed.

Second, what I meant in https://github.com/k0kubun/sqldef/pull/223#issuecomment-1067530640 is to _manually_ remove `AUTO_INCREMENT=xxx` after you use `mysqldef --export` normally. I generally don't expect `--export` output to be used as is. For example, you may want to write comments in `--`. Manually removing `AUTO_INCREMENT=xxx` from `--export` doesn't seem too difficult and I'm not sure why you simply do so instead of changing how sqldef behaves.

Alternatively, I'm okay to remove `AUTO_INCREMENT=xxx` from `--export` output **by default** if there's a reasonable use case that requires this. However, at this point, I think implementation complexity and a tiny risk of users who actually need that outweigh the benefit of it. So I'd like to hear why you cannot just manually remove `AUTO_INCREMENT=xxx` from your schema file after `--export`.

tadamatu:
> However, at this point, I think implementation complexity and a tiny risk of users who actually need that outweigh the benefit of it.

What you are thinking is correct.
After `--export`, it is easy to execute the deletion process with the sed command etc.
I will close this PR.
thank you !

PR #240, opened 2022-05-16
Sort tableNames by table_name

tohae:
`information_schema.tables` is reordered each time the schema is changed, so there is a problem that the difference in the exported schema is large.
To solve this problem, I fixed the order by specifying order by.


k0kubun (approved):
I'm not sure what consistent export output could be useful for, but I have no problem with sorting tables in export output with their names. Thanks!

PR #248, opened 2022-06-15
Supoort posix regexp on psqldef

### LINKED ISSUE #247: [psqldef] Occurs syntax error when POSIX regex in check constraint
hi.

postgresql has regular expressions available in tilde, but psqldef does not seem to support.

## --export output

```
-- No table exists --
```

## Input SQL

```sql
create table t (
  v varchar(1) not null check (v ~ '^\d$')
);
```

## Current output

```sh
found syntax error when parsing DDL "create table t (
  v varchar(1) not null check (v ~ '^\d$')
)": syntax error at position 52
```

## Expected output

```sql
create table t (
  v varchar(1) not null check (v ~ '^\d$')
);
```

ngyuki:
Fix #247


k0kubun (approved):
Nice work :+1:

ngyuki:
Thank you for merge!

Issue #49, opened 2019-11-25
mysql diff does not support boolean?


sonots:
mysqldef version: 0.5.7
mysql version: 5.7.28

With a table like:

```
CREATE TABLE `blahblah` (
  `id` int(11) DEFAULT NULL,
  `feature_flag` char(20) COLLATE utf8_bin DEFAULT NULL,
  `bucket_id` int(11) DEFAULT NULL,
  `name` varchar(100) COLLATE utf8_bin DEFAULT NULL,
  `description` text COLLATE utf8_bin,
  `default_flag` tinyint(1) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
```

mysqldef shows diff everytime as followings and can not be eliminated:

```
ALTER TABLE variations CHANGE COLUMN default_flag default_flag boolean;
```

sonots:
My SQL file is

```
CREATE TABLE IF NOT EXISTS blahblah (
  id INTEGER,
  feature_flag CHAR(20),
  bucket_id INTEGER,
  name VARCHAR(100),
  description TEXT,
  default_flag BOOLEAN
);
```

sonots:
My workaround is to change SQL file as:

```
CREATE TABLE IF NOT EXISTS blahblah (
  id INTEGER,
  feature_flag CHAR(20),
  bucket_id INTEGER,
  name VARCHAR(100),
  description TEXT,
  default_flag TINYINT(1)
);
```

PR #274, opened 2022-08-11
[mysqldef] Fix handling default values of "boolean" correctly.

moajo:
https://github.com/k0kubun/sqldef/issues/49 has been resolved, but not yet when default value is specified.

The specific problem is that it would output meaningless diffs in situations like the test case I added.

The problem was because MySQL treats BOOLEAN values as INT internally, but we were not handling them as such.
see: https://dev.mysql.com/doc/refman/8.0/en/boolean-literals.html

This PR fixes these problems.


k0kubun (approved):
Thank you!

k0kubun:
Released as v0.13.5.

PR #281, opened 2022-09-14
Separate comment schema for each table.

alpaca-tc:
I found a bug when dumping schema on postgresql where column comments are expanded to all tables.
It seems that the process of retrieving comments for the specified table is omitting the specification of the table.

I fixed a problem when retrieving comments by adding a `where` condition to filter by table.


k0kubun (approved):
Thanks!

PR #283, opened 2022-09-28
[mysqldef] Add ssl-ca option 

5st7:
We use RDS as MySQL. 
I want to add an option to specify a server certificate because I want to encrypt the connection to the DB instance with SSL/TLS.

see: https://aws.amazon.com/premiumsupport/knowledge-center/rds-connect-ssl-connection/


k0kubun (approved):
Thank you!

PR #288, opened 2022-10-03
[psqldef] Introduced extension support

alpaca-tc:
`EXTENSION` to be managed in the schema file.

New supported syntax is `CREATE EXTENSION xxx`.
Version checking for extension is not yet supported 🙏 

NOTE: In environments already using any EXTENSION, if you try to apply before adding the extension to the schema file, `DROP EXTENSION xxx` will be applied. Be sure to add `CREATE EXTENSION xxx` to the schema file before applying.


k0kubun (approved):
Thank you!

PR #294, opened 2022-10-21
Add .exe extension to windows executables

### LINKED ISSUE #293: Windows executable file does not have .exe extension
Hi, I just downloaded `sqlite3def_windows_amd64.zip` from the releases page and noticed that the executable file is missing `.exe` extension, thus can't be executed. Renaming `sqlite3def` to `sqlite3def.exe` solved the issue.
I guess there's something wrong in the release workflow and the extension is removed from executables.

ciffelia:
This PR updates Makefile to append `.exe` to windows executables so that they can be executed.

Closes #293

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,17 +8,23 @@ SHELL=/bin/bash
 SQLDEF=$(shell pwd)
 MACOS_VERSION := 11.3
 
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
Thanks!

PR #304, opened 2022-11-10
Add skip_tables option to config yaml

### LINKED ISSUE #303: Proposal: Add skip_tables option to config yaml.
Hi @k0kubun, thank you for the nice plugin!!

I'm using mysqldef with machine learning products, and I have the following approach for databaes.

1. create `tmp_scores` table
2. insert data into `tmp_scores`
3. rename `tmp_scores` to `scores`, and `scores` to `old_scores`
4. delete `old_scores`

This is because the ML data is very huge and we don't want to affect the system using the `scores` table.
So, I would like to add an `ignore_tables` option to the config yaml in order to ignore `tmp_scores`.

If you like this idea, I'll create PR. Thank you.

kitagry:
fix #303


k0kubun (approved):
:+1: Thank you!

PR #306, opened 2022-11-15
[sqlite3def] Add blob type support

knaka:
Added blob type support // [Datatypes In SQLite](https://www.sqlite.org/datatype3.html)

Some virtual table features (such as FTS) create blob columns internally. In that case users cannot avoid using blob types.


cmd/sqlite3def/sqlite3def_test.go
```diff
@@ -143,6 +143,20 @@ func TestSQLite3defConfigIncludesSkipTables(t *testing.T) {
 	assertEquals(t, apply, nothingModified)
 }
 
+func TestSQLite3defBlobType(t *testing.T) {
```
k0kubun, line 146: Can you test this in `TestApply`? All the tests other than `TestApply` should only test CLI features, and SQL manipulation should be tested in `cmd/sqlite3def/tests.yml` because it's faster and less redundant.

PR #308, opened 2022-11-17
sqlite3def: Add support for single-quoted ID

knaka:
Added single-quoted identifier support // [SQLite Keywords](https://www.sqlite.org/lang_keywords.html)

Some virtual table features (such as FTS) create tables with single-quoted IDs. In that case users cannot avoid using them.


k0kubun (approved):
Interesting. Thanks!

PR #311, opened 2022-11-18
Replace get subcommand to install

tomocrafter:
I could not install goyacc on Go 1.19.3 by running `make goyacc` due to command was using `go get`.
This PR replaces go get with go install.

> In Go 1.18, go get will no longer build packages; it will only be used to add, update, or remove dependencies in go.mod. Specifically, go get will always act as if the -d flag were enabled.

https://go.dev/doc/go-get-install-deprecation


k0kubun (approved):
:+1:

PR #314, opened 2022-11-23
Add fs.FS option to sqldef.Run() interface

knaka:
Added an option to `sqldef.Run()` which points to `fs.FS` interface.

Since especially SQLite3 is often used as an embedded database,  it is preferable to be able to read the schema file from the binary-embedded file system rather than from storage.

### DIFF
--- a/sqldef.go
+++ b/sqldef.go
@@ -2,6 +2,7 @@ package sqldef
 
 import (
 	"fmt"
+	"io/fs"
 	"io/ioutil"
 	"log"
 	"os"
@@ -12,6 +13,7 @@ import (
 )
 
 type Options struct {
+	FileSystem  fs.FS
 	DesiredFile string
 	CurrentFile string
 	DryRun      bool
@@ -47,9 +49,18 @@ func Run(generatorMode schema.GeneratorMode, db database.Database, sqlParser dat
 		return
 	}
 
-	sql, err := ReadFile(options.DesiredFile)
-	if err != nil {
-		log.Fatalf("Failed to read '%s': %s", options.DesiredFile, err)
+	var sql string
+	if options.FileSystem == nil {
+		sql, err = ReadFile(options.DesiredFile)
+		if err != nil {
+			log.Fatalf("Failed to read '%s': %s", options.DesiredFile, err)
+		}
+	} else {
+		data, err := fs.ReadFile(options.FileSystem, options.DesiredFile)
+		if err != nil {
+			log.Fatalf("Failed to fs.ReadFile '%s': %s", options.DesiredFile, err)
+		}
+		sql = string(data)
 	}
 	desiredDDLs := sql
 
--- a/cmd/sqlite3def/sqlite3def_test.go
+++ b/cmd/sqlite3def/sqlite3def_test.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	"github.com/k0kubun/sqldef"
 	"github.com/k0kubun/sqldef/cmd/testutils"
 	"github.com/k0kubun/sqldef/database"
 	"github.com/k0kubun/sqldef/database/sqlite3"
@@ -14,6 +15,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"path"
 	"regexp"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


k0kubun:
It's okay to add that since now you maintain it yourself, but may I ask what's the use case for this? Are you integrating sqldef from some Go code?

sqldef is primarily maintained for CLI usages, and we could break API compatibilities for maintenance reasons as long as it doesn't break CLI behaviors. I still don't stop users to use sqldef as a library, but they should own their risk. 

If you care about compatibility, I'd recommend using CLI even from the Go code. If you need to use this API for performance reasons, it'd be nice to see benchmark numbers with and without this PR.

knaka:
As for compatibility, if the parameter `FileSystem` is left nil (not specified), the behavior is the same as before, so code compatibility is preserved.

> may I ask what's the use case for this? Are you integrating sqldef from some Go code?

Yes. Many standalone applications and mobile device applications migrate their built-in database themselves when they are launched. Since programs created in Golang are often provided in a single-binary, I think that some means of embedding the migration information into the binary is necessary especially for SQLite3 support.

Because the only parameter affected is `DesiredFile`, I think it is also valid to pass DDLs as string not using `fs.FS`.

knaka:
> we could break API compatibilities for maintenance reasons as long as it doesn't break CLI behaviors.

Is is okay. As long as you raise the major version number :-)

k0kubun:
> As for compatibility, if the parameter FileSystem is left nil (not specified), the behavior is the same as before, so code compatibility is preserved.

Oh, I mean, I'm not complaining about compatibility issues in this PR. I just wanted to inform you that any Go code using sqldef as a library could be broken at any version upgrade since we don't care about API compatibilities. Again, this PR is fine.

> Is is okay. As long as you raise the major version number :-)

I'm exactly talking about this. sqldef's versioning only cares about CLI, and the maintainers reserve the right to change API at any version upgrade as long as it doesn't change CLI behaviors. It's very possible that a teeny version upgrade breaks your application, although it's kind of rare in reality. If you can just ignore versioning, it should be fine.

> Many standalone applications and mobile device applications

Just curious, instead of talking about general problems, could you talk about your case? Do you have a Go server-side application that uses SQLite3, a mobile application using SQLite3 written in Go, or both?

sqldef.go
```diff
@@ -47,9 +49,18 @@ func Run(generatorMode schema.GeneratorMode, db database.Database, sqlParser dat
...
-	if err != nil {
-		log.Fatalf("Failed to read '%s': %s", options.DesiredFile, err)
+	var sql string
+	if options.FileSystem == nil {
+		sql, err = ReadFile(options.DesiredFile)
+		if err != nil {
+			log.Fatalf("Failed to read '%s': %s", options.DesiredFile, err)
+		}
+	} else {
+		data, err := fs.ReadFile(options.FileSystem, options.DesiredFile)
```
k0kubun, line 59: Could it be possibly simpler to always use `fs.ReadFile` by providing something in the first argument when `options.FileSystem == nil`? I'm not sure what kind of `fs.ReadFile(x, y)` works like `ReadFile(y)`, but it might be nice if we can avoid complicating the implementation by this if-else branch.

knaka:
> Oh, I mean, I'm not complaining about compatibility issues in this PR. 

Sorry, it is my misunderstanding.

> sqldef's versioning only cares about CLI, and the maintainers reserve the right to change API at any version upgrade as long as it doesn't change CLI behaviors.

I understand it. Compared to some CLI projects which cannot be linked because the code is in `internal` directory, the fact that sqldef is linkable is very nice.

> Just curious, instead of talking about general problems, could you talk about your case? Do you have a Go server-side application that uses SQLite3, a mobile application using SQLite3 written in Go, or both?

I am maintaining a desktop application which is written in Wails -  a framework for Golang that works similar to Electron. // [The Wails Project | Wails](https://wails.io/)

Now I am using go-migrate for the application while planning to replace it with sqldef. The application applies necessary DB migration scripts when it is launched.

Of course, my server-side applications are using MySQL and PostgreSQL and CLI must be better to do DB migration than embedding it to binary.

k0kubun:
> I am maintaining a desktop application which is written in Wails - a framework for Golang that works similar to Electron. // [The Wails Project | Wails](https://wails.io/)

Very interesting! Great use case of sqldef :heart_eyes:

k0kubun (approved):
I'd like to support your project, so please feel free to merge this no matter how https://github.com/k0kubun/sqldef/pull/314#discussion_r1031102154 goes. If you think my proposal rather complicates the implementation, it's fine.

knaka:
I am sorry that you have approved, let me close this pull request. I would like to adopt a simpler way (#315).

I had a stereotype that because go-migrate used fs.FS, I had to use the same way, but I realized that I did not have to. while go-migrate has to manage a series of many migration scripts, sqldef has to parse only one set of DDLs.

PR #315, opened 2022-11-24
Add DesiredDDLs option to pass DDLs as string

knaka:
Added option to `sqldef.Run()` to pass DDLs as a string so that sqldef can be used in single binary without specifying schema file on the filesystem.

At first I was going to use `fs.FS` (#314) , but then I decided it was unnecessary, so I created this pull request instead. See the PR  for use cases which requires this this fix and the fear that interface changes will break code compatibility.

### DIFF
--- a/cmd/mssqldef/mssqldef.go
+++ b/cmd/mssqldef/mssqldef.go
@@ -53,8 +53,17 @@ func parseOptions(args []string) (database.Config, *sqldef.Options) {
 	}
 
 	desiredFile, currentFile := sqldef.ParseFiles(opts.File)
+
+	var desiredDDLs string
+	if !opts.Export {
+		desiredDDLs, err = sqldef.ReadFile(desiredFile)
+		if err != nil {
+			log.Fatalf("Failed to read '%s': %s", desiredFile, err)
+		}
+	}
+
 	options := sqldef.Options{
-		DesiredFile: desiredFile,
+		DesiredDDLs: desiredDDLs,
 		CurrentFile: currentFile,
 		DryRun:      opts.DryRun,
 		Export:      opts.Export,
--- a/sqldef.go
+++ b/sqldef.go
@@ -12,7 +12,6 @@ import (
 )
 
 type Options struct {
-	DesiredFile string
 	DesiredDDLs string
 	CurrentFile string
 	DryRun      bool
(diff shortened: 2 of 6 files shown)


sqldef.go
```diff
@@ -13,6 +13,7 @@ import (
 
 type Options struct {
```
k0kubun, line 14: Can we just let it have only `DesiredDDLs` and `CurrentDDLs`? Having both `DesiredFile` and `DesiredDDLs` seems confusing.

knaka: Trying to make a diff as small as possible is a bad habit of mine.

sqldef.go
```diff
@@ -13,6 +13,7 @@ import (
...
 	DesiredFile string
```
k0kubun, line 15:
```suggestion
```

In case it wasn't clear, at https://github.com/k0kubun/sqldef/pull/315#discussion_r1031861301, I suggested to remove this field. The caller should be able to read the `DesiredFile`, and pass that as `DesiredDDLs`. Having both in `Options` seems redundant and confusing. It's nicer to make unexpected situations simply not compilable instead of making it a runtime error.

knaka: I made a misunderstanding. You are right.

knaka: Each `*sql*def.go` has become somewhat redundant. What do you think?

cmd/mssqldef/mssqldef.go
```diff
@@ -53,8 +53,17 @@ func parseOptions(args []string) (database.Config, *sqldef.Options) {
 	}
 
 	desiredFile, currentFile := sqldef.ParseFiles(opts.File)
+
+	var desiredDDLs string
+	if !opts.Export {
+		desiredDDLs, err = sqldef.ReadFile(desiredFile)
```
k0kubun, line 59: Given that we resort to `log.Fatalf` anyway, you could call `log.Fatalf` in `sqldef.ReadFile` and only return a string. Then you should also be able to do `options.DesiredDDLs = sqldef.ReadFile(desiredFile)` after defining `options` so that you can skip defining an extra local variable `desiredDDLs`.

k0kubun (approved):
:+1: Please feel free to merge this. It's not difficult to fix https://github.com/k0kubun/sqldef/pull/315#discussion_r1033007388 at any time.

cmd/mssqldef/mssqldef.go
knaka: `sqldef.ReadFile` is also used in other parts of the code so I would like to fix it in another PR.

PR #319, opened 2022-11-26
Change index types to upper case in ALTER TABLE

tagomoris:
For the consistency between `ALTER TABLE ... DROP KEY` and `ALTER TABLE ... ADD key`, etc.

### DIFF
--- a/cmd/mssqldef/mssqldef_test.go
+++ b/cmd/mssqldef/mssqldef_test.go
@@ -497,7 +497,7 @@ func TestMssqldefCreateTableAddPrimaryKeyConstraint(t *testing.T) {
 	)
 
 	assertApplyOutput(t, createTable, applyPrefix+
-		"ALTER TABLE [dbo].[users] ADD CONSTRAINT [pk_users] primary key CLUSTERED ([id] desc);\n",
+		"ALTER TABLE [dbo].[users] ADD CONSTRAINT [pk_users] PRIMARY KEY CLUSTERED ([id] desc);\n",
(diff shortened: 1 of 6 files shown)


tagomoris:
Changing `primary key` into `PRIMARY KEY` in schema/ast.go only affects the case when `PRIMARY KEY` is specified on column definition lines. All others still have `primary key`.
* https://github.com/k0kubun/sqldef/actions/runs/3552934014/jobs/5968153375#step:6:45
* https://github.com/k0kubun/sqldef/actions/runs/3552934014/jobs/5968153405#step:6:20

The `index.indexType` [here](https://github.com/k0kubun/sqldef/blob/808598e5e2a52b18b456e37bc99d8865d6dffb0d/schema/generator.go#L1108-L1112) is derived from `indexDef.Info.Type` [here](https://github.com/k0kubun/sqldef/blob/808598e5e2a52b18b456e37bc99d8865d6dffb0d/schema/parser.go#L177), and it's set by yacc-generated code [for example, here](https://github.com/k0kubun/sqldef/blob/master/parser/parser.go#L6680). I guess it contains lower case characters.

So, in my idea, the only possible solution is to `strings.ToUpper(index.indexType)` in generator.go.

tagomoris:
Should I clean up or squash commits?

k0kubun:
I use "Squash and merge" by default, so it's okay :slightly_smiling_face:

k0kubun (approved):
Thank you for looking into this!

PR #325, opened 2022-12-04
Enable to set multiple files and change desired file option

### LINKED ISSUE #322: Proposal: change current behavior for --file option
Hi @k0kubun. Thank you for the nice tool!

I would like to use the following schema.

```
./schema
├── common.sql  # for both production and test
├── production.sql
└── test.sql
```

I think it is natural to use `--file` option multiple, like the following.

```bash
mysqldef --file ./schema/common.sql --file ./schema/production.sql
```

But, currently `--file` option has other role. https://github.com/k0kubun/sqldef/pull/179.
I think it is a bit confusing, because file options have a different role and there is no document.

```diff
$ mysqldef --help
...
-      --file=sql_file                Read schema SQL from the file, rather than stdin (default: -)
+      --file=sql_file                Read schema SQL from the files, rather than stdin (default: -)
+      --current-file=sql_file        Read schema SQL file as current schema, rather than to connect DB
``` 

I would like to change the current option like above.
But, it is breaking change. So if you don't like this idea, please feel free to close this issue. Thank you.

kitagry:
fix #322 

## What changed

```bash
# We can set multiple files option
$ mysqldef --file a.sql,b.sql

# We can set desired file, rather than connect db
$ mysqldef current.sql < desired.sql
```


sqldef_test.go
```diff
@@ -0,0 +1,37 @@
+package sqldef_test
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/k0kubun/sqldef"
+)
+
+func TestParseFiles(t *testing.T) {
```
k0kubun, line 10: Could you write CLI tests only at a `cmd/*def/*def_test.go` level? For sqldef, the only thing that matters in terms of backward compatibility is CLI, and any functions used for implementing it are just implementation details. When you want to refactor it in a way that modifies `sqldef.ParseFiles`, you will need to modify this test as well, but it'll be an unneeded maintenance cost when you don't need to maintain this function's compatibility.

sqldef.go
```diff
@@ -68,22 +69,32 @@ func Run(generatorMode schema.GeneratorMode, db database.Database, sqlParser dat
...
-	desiredFile := files[0]
-	currentFile := ""
-	if len(files) == 2 {
-		desiredFile = files[1]
-		currentFile = files[0]
-	} else if len(files) > 2 {
-		fmt.Printf("Expected only one or two --file options, but got: %v\n", files)
-		os.Exit(1)
+	result := make([]string, 0, len(files))
+	re := regexp.MustCompile("[, ]")
```
k0kubun, line 78: Which are you going to actually use, `,` or ` `? If you're going to use only either of these, nobody might use the other, so I want to support only one of these.

kitagry: I'll use `,`! I'll delete ` `.

k0kubun (approved):

kitagry:
Thank you for the review!

PR #329, opened 2022-12-09
mssqldef: Added support for datetime2 with precision option.

odz:


cmd/mssqldef/mssqldef_test.go
```diff
@@ -61,6 +61,7 @@ func TestMssqldefColumnLiteral(t *testing.T) {
 		  v_money money,
 		  v_datetimeoffset datetimeoffset(1),
 		  v_datetime2 datetime2,
+		  v_datetime2_with_precision datetime2(3),
```
k0kubun, line 64: I'm trying to move any new/modified tests to `cmd/mssqldef/tests.yml`. It's not your fault that this test is still here, but could you move this test to `cmd/mssqldef/tests.yml`?

odz:
Thank you for review. I've fixed a test.

k0kubun (approved):
Thank you :+1:

PR #335, opened 2022-12-13
psqldef: Parse interval type

### LINKED ISSUE #334: Syntax error when parsing INTERVAL type on postgres
## --export output

psqldef 3a74b86 (v0.15.1)

```
2022/12/13 14:51:09 found syntax error when parsing DDL "CREATE TABLE "public"."foo" (
    "bar" interval
)": syntax error at position 49 near 'interval'
```

## Input SQL

```
CREATE TABLE "public"."foo" (
    "bar" interval
);
```

iwamatsu0430:
fixes #334 

I fixed parser for `interval` type.
If there is better way, please let me know.


k0kubun (approved):
Thank you!

PR #350, opened 2023-01-23
[mysqldef] Support window function on mysql.

keisuke-matsufuji:
The following clause have been added to allow the Window function 
to be used in Mysql when implementing View tables.

- Aggregation using the OVER clause
- Using "PARTITION BY" and "ORDER BY" in the OVER clause


k0kubun (approved):
Thanks!

PR #331, opened 2022-12-10
Make able to alter add constraint after create table

tomocrafter:
fixes #327 

I implemented this following another statements like `add policy`.
If it is incorrect and there is better option to implement it, please let me know.

Thank you.

PR #352, opened 2023-01-30
accept ALTER TABLE table_name ADD foreign_key_definition (without ONLY)

fumieval:
This change allows `mysqldef` to accept `ALTER TABLE ... ADD FOREIGN KEY` statement, not just `ALTER TABLE ONLY` which has been supported by #331 . IIUC this is needed to be compatible with MySQL.


k0kubun (approved):
Thanks!

PR #354, opened 2023-02-06
[mysqldef] Using window function on replace view.

keisuke-matsufuji:
see https://github.com/k0kubun/sqldef/pull/350

a syntax error was occurring when using the window function in a replace view statement.

modified window function can be used in replace view statement


k0kubun (approved):
Thanks!

Issue #301, opened 2022-11-07
support MySQL 8.0 Generated Column


tomtwinkle:
Looking forward to support for Generated columns!
https://dev.mysql.com/doc/refman/8.0/en/create-table-generated-columns.html

## Platform

* OS: Linux
* RDBMS: MySQL
* Version: 8.0

## --export output

```
CREATE TABLE `samples` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `json_data` json NOT NULL,
  `json_name` varchar(20) GENERATED ALWAYS AS (json_extract(`json_data`,'$.foo_name')) VIRTUAL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

## Input SQL

Changes in the contents of columns referenced in GENERATED. `$.foo_name` -> `$.bar_name`.

```
CREATE TABLE `samples` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `json_data` JSON NOT NULL,
  `json_name` varchar(20) GENERATED ALWAYS AS (JSON_EXTRACT(`json_data`, "$.bar_name")) VIRTUAL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

## Current output

No change

## Expected output

```
ALTER TABLE `samples` DROP `json_name`;
ALTER TABLE `samples` ADD `json_name` varchar(20) GENERATED ALWAYS AS (JSON_EXTRACT(`json_data`, "$.bar_name")) VIRTUAL;
```

k0kubun:
Can you write an example schema SQL you'd like to use? It's okay you didn't follow the issue template, but still, I always want to see an actual SQL instead of a reference to general features so that we don't need to support something that nobody uses.

tomtwinkle:
Sorry.
I have included the details. Please check it.

k0kubun:
Thanks for the update. Since sqldef seems to be parsing that syntax correctly, it might not be too difficult to fix. Part of [schema/generator.go](https://github.com/k0kubun/sqldef/blob/master/schema/generator.go) may be not detecting that kind of change. Would you be interested in working on a PR to fix it?

tomtwinkle:
Thanks, I'll take a look.

PR #355, opened 2023-02-06
[mysqldef] support MySQL 8.0 Generated Column

tomtwinkle:
https://github.com/k0kubun/sqldef/issues/301

Support for Generated Column

For now, only MySQL is supported.

### DIFF
--- a/schema/generator.go
+++ b/schema/generator.go
@@ -1559,6 +1559,7 @@ func (g *Generator) areSameGenerated(generatedA, generatedB *Generated) bool {
 	if generatedA == nil || generatedB == nil {
 		return false
 	}
+	// TODO: Difference between bracketed and unbracketed, as Expr values are not fully comparable.
 	return (generatedA.expr == generatedB.expr || generatedA.expr == "("+generatedB.expr+")") &&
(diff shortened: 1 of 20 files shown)


schema/generator.go
```diff
@@ -1512,6 +1552,18 @@ func (g *Generator) haveSameColumnDefinition(current Column, desired Column) boo
...
+func (g *Generator) areSameGenerated(generatedA, generatedB *Generated) bool {
+	if generatedA == nil && generatedB == nil {
+		return true
+	}
+	if generatedA == nil || generatedB == nil {
+		return false
+	}
+	// TODO: Difference between bracketed and unbracketed, as Expr values are not fully comparable.
+	return (generatedA.expr == generatedB.expr || generatedA.expr == "("+generatedB.expr+")") &&
+		generatedA.generatedType == generatedB.generatedType
```
tomtwinkle, line 1564: TODO
Defining a lighter expression without parentheses, such as A / B * C instead of (A / B) * C in Expr, makes a difference every time.

k0kubun (approved):
Thanks!

PR #356, opened 2023-02-18
[mssqldef] support nonclustered columnstore index

odz:
Support NonClustered ColumnStore INDEX in SQL Server.


k0kubun (approved):
Thank you!

PR #367, opened 2023-03-06
[mssqldef] Add semicolon to exported view definition.

### LINKED ISSUE #366: mssqldef: Operation fails when there are multiple views in database.
## Overview

If there are two or more views in database, any operations will fail.

## Platform

* OS: Linux (probably all)
* RDBMS: SQL Server
* Version: 0.15.13

## Current schema

NOTE: `--export` operation fails.

```sql
CREATE VIEW v1 AS SELECT 1 AS N;
CREATE VIEW v2 AS SELECT 2 AS N;
```

## Input SQL

```sql
CREATE VIEW v1 AS SELECT 1 AS N;
```

## Current output

```
2023/03/07 00:42:17 found syntax error when parsing DDL "CREATE VIEW v1 AS SELECT 1 AS N

create view v2 as select 2 as N": syntax error at position 40 near 'create'
```

## Expected output

```sql
DROP VIEW [dbo].[v2];
```

odz:
fix #366

### DIFF
--- a/cmd/mssqldef/tests.yml
+++ b/cmd/mssqldef/tests.yml
@@ -133,3 +133,10 @@ DropTableOnNonStandardDefaultSchema:
   output: |
     DROP TABLE [FOO].[v];
   user: "mssqldef_user"
+MultipleView:
+  current: |
+    CREATE VIEW v1 AS SELECT 1 AS N;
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
Thanks!

PR #370, opened 2023-03-13
MySQL's default index is B-Tree, so if not specified, the comparison is performed as B-Tree

tomtwinkle:
Now, there is a difference between Index unspecified and USING BTREE.
fixed to set default value when comparing Indexes so that there is no difference.

* RDBMS: MySQL
* Version: -

## --export output
```
CREATE TABLE users (
  id int(11) NOT NULL,
  KEY index_id (id)
);
```

## Input SQL
```
CREATE TABLE users (
  id int(11) NOT NULL,
  KEY index_id (id) USING BTREE
);
```

## Now output

```
ALTER TABLE `users` DROP KEY `index_id` (`id`);
ALTER TABLE `users` ADD KEY `index_id` (`id`) USING BTREE;
```

## This PR output

```
-- Nothing is modified --
```


k0kubun (approved):
Thanks!

Issue #375, opened 2023-03-19
Support for changing timezone in psqldef


mkusaka:
Thanks for this great product!

psqldef does not seem to support changing the timezone of a column of type timestamp/time.

## Platform

* OS: macOS
* RDBMS: PostgreSQL
* Version: v0.15.18

## --export output
```sql
CREATE TABLE "public"."test" (
    "id" bigserial NOT NULL,
    "created_at" timestamp with time zone NOT NULL,
    PRIMARY KEY ("id")
);
```

## Input SQL

```sql
CREATE TABLE public.test (
  "id" bigserial NOT NULL,
  "created_at" timestamp without time zone NOT NULL,
  PRIMARY KEY ("id")
);
```

## Current output

```
-- Nothing is modified --
```

## Expected output

```sql
ALTER TABLE "public"."test" ALTER COLUMN created_at TYPE timestamp without time zone;
```

k0kubun:
Thanks for filing this. Would you be interested in working on a PR to fix this? I think this can be fixed in `schema/generator.go`, maybe in `haveSameDataType`. Once fixed, it'd be nice to have a test in `cmd/psqldef/tests.yml` as well.

mkusaka:
@k0kubun Thank you for letting me know what changes need to be made.
I have just created a PR of the changes. :)
https://github.com/k0kubun/sqldef/pull/376

PR #376, opened 2023-03-20
[psqldef] support for changing timezone in time/timestamp column

mkusaka:
fix https://github.com/k0kubun/sqldef/issues/375


k0kubun (approved):
Thanks!

PR #379, opened 2023-03-20
add postgresql 15 to test matrix

mkusaka:
posgtresql v15 was released, so I added to the test matrix.

https://www.postgresql.org/docs/15/release-15.html#id-1.11.6.7.4
From v15, postgres remove PUBLIC creation permission on the public schema as default.
So, I added GRANT ALL to dummy role for test.


k0kubun (approved):
:+1:

PR #398, opened 2023-05-11
Escape parameters in unique constraint query for postgres

omochi:
# Issue

`psqldef` fails to export when database has unique index whose name needs escaping.

```
$ psqldef -U postgres work --export
2023/05/11 16:37:49 syntax error at or near ":"
```

In my case, I faced it with the name `uq:_fluent_migrations.name`.

Here is screenshot of the table.

スクリーンショット 2023-05-11 16 37 19

This is created by Fluent (https://github.com/vapor/fluent) 
which is the most popular web framework for Swift 😄

# Patch

This patch fixes the issue.

I added test case.
It fails on master.

```
=== RUN   TestApply/EscapeUniqueIndexQuery
    testutils.go:121: syntax error at or near ":"

FAIL
FAIL	github.com/k0kubun/sqldef/cmd/psqldef	85.223s
FAIL
```

It success on this patch.

### DIFF
--- a/cmd/psqldef/tests.yml
+++ b/cmd/psqldef/tests.yml
@@ -472,6 +472,22 @@ IndexesOnChangedExpressions:
   output: |
     DROP INDEX "public"."function_index";
     CREATE UNIQUE INDEX function_index ON public.test (jsonb_extract_path_text(col, 'foo'));
+EscapeUniqueIndexQuery:
+  current: |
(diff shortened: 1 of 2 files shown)


omochi:
I fixed other test cases.
They are failed since patch changed output to escaped.
I updated expectation of these tests.

k0kubun (approved):
Thank you!

Issue #292, opened 2022-10-16
I want to change option `--skip-drop` to `--enable-drop`


masak1yu:
Thank you for the marvelous tool!
Currently, if used without any options, some diffs will drop the table.
If not careful with this specification, there is a risk of unintended table deletion.
As a request, I would like you to change so that it skips dropping tables without options and only drops tables with options.

```
before
--skip-drop

after
--enable-drop
```

f.y.i
https://github.com/ridgepole/ridgepole/issues/105

Please confirm.

k0kubun:
Sounds good to me. 

Should the option target only tables, or do you want to skip dropping indexes, constraints, columns, and everything else by default too? I wonder whether what you really want is `--enable-drop-table` like ridgepole (disabling only `DROP TABLE` by default) or just `--enable-drop` (disable any `DROP` DDLs by default).

msky026:
I think, Skipping all the latter all DDL seems to me more reasonable in production.
Therefore, I think the option is `--enable-drop`.

k0kubun:
OK. Please feel free to file a patch that adds `--enable-drop` to sqldef commands. I'll accept it.

> I think, Skipping all the latter all DDL seems to me more reasonable in production.

Just to ground your design decision, can you provide a more concrete use case? Here's an example use case, which is a counter-argument to your claim (I'm not saying it should be `--enable-drop-table` here, but at least the following text seems more compelling than what you've just said :slightly_smiling_face:):

---

Let's say you have only `users` table in development and CI. You have `users` and `users_backup` in production. When you start using sqldef, you added only `users` table because you can only see that in development and CI. Then, you tried applying the schema to production. You managed to notice that it's going to `DROP TABLE users_backup`, but you felt scared that it could drop such tables. Dropping tables is a fairly rare operation, so you're okay with making it disabled by default.

Now, you found that `users` table has an unnecessary index that used to be in use by a query that no longer exists. To speed up INSERT, you want to drop the index. You removed the index definition from your schema.sql. CI automation executed sqldef, but it wouldn't delete the index if any DROP is disabled by default. Unlike `users_backup`, you're pretty much aware of the `users` table and everything you have in your schema.sql. So when you remove the index, you mean it. There's no circumstance where you delete a line of an index when you don't want to delete the index.

msky026:
In the above example, it seems that there is little reason to suppress the drop for index.
(I think the same is true for constraints)
[...]

PR #399, opened 2023-05-11
change option skip drop to enable drop

msky026:
https://github.com/k0kubun/sqldef/issues/292

It changed option that skip-drop to enable-drop.
Only tables will be dropped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Application Options:
       --file=sql_file               Read schema SQL from the file, rather than stdin (default: -)
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
-      --skip-drop                   Skip destructive changes such as DROP
+      --enable-drop                 Ebale destructive changes such as DROP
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ Application Options:
       --file=sql_file               Read schema SQL from the file, rather than stdin (default: -)
       --dry-run                     Don't run DDLs but just show them
       --export                      Just dump the current schema to stdout
-      --skip-drop                   Skip destructive changes such as DROP
+      --enable-drop                 Ebale destructive changes such as DROP
```
k0kubun, line 38: I'm sorry that my comments in https://github.com/k0kubun/sqldef/issues/292 were confusing, but can we call it `--enable-drop-table`? It'd be difficult to correctly guess the actual behavior from `--enable-drop`.

It'd be also nice to update the description on the right column accordingly. `Ebale` looks like a typo.

msky026: I think so.
I modify `enable-drop` to `enable-drop-table`.

k0kubun (approved):
Thanks! Let's try this.

PR #400, opened 2023-05-18
Support casting a default value to array

hokaccha:
This change fixes an error that occurred when the default value had an array typecast.

```sql
CREATE TABLE tbl (
  a INT[] NOT NULL DEFAULT '{}'::INT[]
);
```

```
$ psql --host localhost --user postgres sandbox < schema.sql 
CREATE TABLE

$ psqldef --export --host localhost --user postgres sandbox    
2023/05/18 15:47:01 found syntax error when parsing DDL "CREATE TABLE "public"."tbl" (
    "a" integer[] NOT NULL DEFAULT '{}'::integer[]
)": syntax error at position 80
```


k0kubun (approved):
Thank you!

PR #401, opened 2023-05-20
[psqldef]Fix Error in Materialized Views with Multiple Indices

yudppp:
## ISSUE

There was a bug causing an error when multiple indices existed in a MATERIALIZED VIEW or when multiple MATERIALIZED VIEWs were present.

exmple

schema.sql
```sql
CREATE MATERIALIZED VIEW public.view_users AS SELECT * FROM users;
CREATE INDEX index_name on public.view_users (name);
CREATE UNIQUE INDEX index_age on public.view_users (age);
```

get ddl
```sql
CREATE MATERIALIZED VIEW public.view_users AS SELECT * FROM users;
CREATE INDEX index_name on public.view_users (name)
CREATE UNIQUE INDEX index_age on public.view_users (age)
```

## PATCH

The reason was the absence of a semicolon at the end of the index. This time, we have addressed the issue by adding semicolons and also added test codes.

### DIFF
--- a/cmd/psqldef/psqldef_test.go
+++ b/cmd/psqldef/psqldef_test.go
@@ -568,7 +568,34 @@ func TestPsqldefCreateView(t *testing.T) {
 	}
 }
 
+func TestPsqldefCreateMaterializedView(t *testing.T) {
+	for _, tc := range publicAndNonPublicSchemaTestCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			resetTestDatabase()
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
I wish we could write new tests in `cmd/psqldef/tests.yml`, but I guess the `mustExecuteSQL` part is not possible for now. Thank you for fixing the bug!

yudppp:
thanks!

PR #402, opened 2023-05-20
Enhance postgres parser tests by comparing with generic parser results

hokaccha:
In this pull request, we have improved the postgres parser tests by comparing the parsing results with those of the generic parser.

Previously, the tests only checked if the parsing process completed without errors, which made it difficult to verify if the parsing was done correctly. 

To resolve this issue, I suggest comparing the postgres parser test results with the generic parser results to ensure they are identical. However, due to the complexity of making every test case match the generic parser's results, the comparison is done only for cases with the `compare_with_generic_parser` option enabled.

To perform the struct comparisons, I used the [stretchr/testify](https://github.com/stretchr/testify) library, which provided the most desirable behavior after testing other libraries such as [google/go-cmp](https://github.com/google/go-cmp) and [go-test/deep](https://github.com/go-test/deep). If the test fails, the output is as followings:

```
    parser_test.go:38:
                Error Trace:    /Users/hokaccha/local/src/github.com/k0kubun/sqldef/database/postgres/parser_test.go:38
                Error:          Not equal:
                                expected: []database.DDLStatement{database.DDLStatement{DDL:"CREATE TABLE public.bigdata (\n  data bigint\n)",
Statement:(*parser.DDL)(0x140000ecd20)}}
                                actual  : []database.DDLStatement{database.DDLStatement{DDL:"CREATE TABLE public.bigdata (\n  data bigint\n)",
Statement:(*parser.DDL)(0x140000ecb40)}}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -72,4 +72,3 @@
                                     Checks: ([]*parser.CheckDefinition) <nil>,
                                -    Options: (map[string]string) {
                                -    }
                                +    Options: (map[string]string) <nil>
                                    }),
                Test:           TestParse/CreateTableWithSchema
```

### DIFF
--- a/database/postgres/parser.go
+++ b/database/postgres/parser.go
@@ -222,7 +222,8 @@ func (p PostgresParser) parseSelectStmt(stmt *pgquery.SelectStmt) (parser.Select
 		SelectExprs: selectExprs,
 		From: parser.TableExprs{
 			&parser.AliasedTableExpr{
-				Expr: fromTable,
+				Expr:       fromTable,
+				TableHints: []string{},
 			},
 		},
 	}, nil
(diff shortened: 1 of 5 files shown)


k0kubun (approved):
Thanks!

PR #403, opened 2023-05-21
Support for update comment

hokaccha:
I fixed an issue where the difference was not output when the comment was changed.


k0kubun (approved):
:+1:

PR #404, opened 2023-05-22
Support alter table constraint for postgres parser

hokaccha:
Added support for constraint syntax to add unique index and foreign key in ALTER TABLE. The following SQL can now be parsed by the postgres parser.

```sql
ALTER TABLE users ADD CONSTRAINT username UNIQUE (name, age);
ALTER TABLE ONLY public.posts ADD CONSTRAINT posts_ibfk_1 FOREIGN KEY (user_id) REFERENCES users (id) ON DELETE SET NULL ON UPDATE RESTRICT;
```


k0kubun (approved):
Thank you!

PR #408, opened 2023-05-25
Fix errors caused by updating to pgquery v4

hokaccha:
The changes on https://github.com/k0kubun/sqldef/pull/404 were not compatible with the type updated on https://github.com/k0kubun/sqldef/pull/406.


k0kubun (approved):
:+1:

Issue #397, opened 2023-05-09
[psqldef] Cannot parse `t.*` in `select t.* from t;` 


halwhite:
## Platform
In version 0.11.50 (that we are using now), Input SQL is loaded correctly, but in version 0.15.24, an error occurs when loading.

* OS: macOS
* RDBMS: PostgreSQL
* Version: 0.15.24

## --export output
```
-- No table exists --
```

## Input SQL
```
CREATE TABLE
  users (
    usre_id INT PRIMARY KEY,
    name VARCHAR(100) NOT NULL
  );

CREATE VIEW
  users_view AS
SELECT
  u.*
FROM
  users AS u;
```

## Current output

```
panic: interface conversion: pg_query.isNode_Node is *pg_query.Node_AStar, not *pg_query.Node_String_

goroutine 1 [running]:
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseExpr({{0x0?}, 0x0?}, 0x0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:330 +0x8c4
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseResTarget({{0x0?}, 0x0?}, 0x400009cba0)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:231 +0x28
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseSelectStmt({{0x1?}, 0x0?}, 0x40001c0000)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:198 +0x1a8
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseViewStmt({{0x20?}, 0x40?}, 0x40000f5f80)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:169 +0x90
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseStmt({{0x40003160f7?}, 0x60?}, 0x40001d7701?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:76 +0x120
github.com/k0kubun/sqldef/database/postgres.PostgresParser.Parse({{0x200?}, 0x1?}, {0x40003160a0, 0x95})
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:45 +0x14c
[...]

PR #423, opened 2023-07-16
Support `select *` for postgres parser

hokaccha:
This PR resolves the error in https://github.com/k0kubun/sqldef/issues/397. However, there is a problem as follows.

```sql
# schema.sql
CREATE TABLE public.test_table (
  "col1" numeric,
  "col2" numeric
);
CREATE VIEW public.test_view AS SELECT t.* FROM test_table t;
```

```
$ go run ./cmd/psqldef -h localhost -U postgres sandbox < schema.sql
-- Apply --
CREATE TABLE public.test_table (
  "col1" numeric,
  "col2" numeric
);
CREATE VIEW public.test_view AS SELECT t.* FROM test_table t;
```

```
# Run the same command again
$ go run ./cmd/psqldef -h localhost -U postgres sandbox < schema.sql
-- Apply --
CREATE OR REPLACE VIEW "public"."test_view" AS select t.* from test_table as t;
```

No matter how many times `psqldef` is executed, `CREATE OR REPLACE VIEW` is executed. It seems that the cause is that when `select *` is specified, the query is saved as a query with expanded columns.

```
$ psql -h localhost -U postgres sandbox -c "SELECT definition FROM pg_views WHERE viewname = 'test_view'"
      definition
-----------------------
  SELECT t.col1,      +
     t.col2           +
    FROM test_table t;
(1 row)
```

Therefore, it becomes as follows:

- `currentDDL`: `SELECT t.col1, t.col2`
- `desiredDDL`: `SELECT t.*`

As a result, the comparison fails. This seems to be the same not only for postgres but also for mysql. I have confirmed that the same phenomenon occurs with `mysqldef`.

To solve this problem, we need a process to expand `*` when parsing `desiredDDL`. This is not impossible, but it seems to take time. Another option is to have sqldef output an error that it does not support `*` and exit until this problem is solved.

@k0kubun Which do you think is better?

### DIFF
--- a/database/postgres/parser.go
+++ b/database/postgres/parser.go
@@ -209,6 +209,7 @@ func (p PostgresParser) parseSelectStmt(stmt *pgquery.SelectStmt) (parser.Select
 	}
 
 	var fromTable parser.TableName
+	var aliasName string
 	if len(stmt.FromClause) == 0 {
 		fromTable = parser.TableName{
 			Name:      parser.NewTableIdent(""),
@@ -222,6 +223,9 @@ func (p PostgresParser) parseSelectStmt(stmt *pgquery.SelectStmt) (parser.Select
 			if err != nil {
 				return nil, err
 			}
+			if node.RangeVar.Alias != nil {
+				aliasName = node.RangeVar.Alias.Aliasname
+			}
 		default:
 			return nil, fmt.Errorf("unknown node in parseSelectStmt: %#v", node)
 		}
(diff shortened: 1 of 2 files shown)


k0kubun:
> To solve this problem, we need a process to expand * when parsing desiredDDL. This is not impossible, but it seems to take time. Another option is to have sqldef output an error that it does not support * and exit until this problem is solved.

For the short term, it makes sense to give up comparison and assume nothing has changed when one side has `select *`. Eventually, we should compare them at an AST level, normalizing `*` using the table schema in the previous DDLs. But let's take a workaround for now.

And, thanks for fixing the parser. This change itself is innocent, so let me just merge this.

PR #424, opened 2023-07-17
Fix the description of `enable-drop-table` option

y-yagi:
I'm sorry if I misunderstood something, but this option enables `DROP`, so `skip` isn't the correct description I think.


k0kubun (approved):
Thanks!

PR #425, opened 2023-07-20
Don't use deprecated `ioutil` package

y-yagi:
`ioutil` package is deprecated since Go 1.16.
https://pkg.go.dev/io/ioutil#pkg-overview

This fixes not to use a function of `ioutil`.


k0kubun (approved):
:+1:

PR #431, opened 2023-08-20
Migrate from deprecated ioutil to compatible functions

hokaccha:
Replaced the `ioutil` package with compatible functions as `ioutil` package is deprecated.

https://pkg.go.dev/io/ioutil

> Deprecated: As of Go 1.16, the same functionality is now provided by package [io](https://pkg.go.dev/io) or package [os](https://pkg.go.dev/os), and those implementations should be preferred in new code.

### DIFF
--- a/cmd/testutils/testutils.go
+++ b/cmd/testutils/testutils.go
@@ -3,8 +3,8 @@ package testutils
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
+	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


k0kubun (approved):
:+1: Thanks! Please feel free to merge/release pull requests without waiting for my reviews, especially when you don't have anything to discuss or carefully review.

hokaccha:
ok, thanks!

PR #432, opened 2023-08-31
Support function call as default expression

cockscomb:
This PR allows us to specify not only literal constants but also function call expressions in the default clause for columns.

- [MySQL :: MySQL 8.0 Reference Manual :: 11.6 Data Type Default Values](https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#data-type-defaults-explicit)
- [PostgreSQL: Documentation: 15: 5.2. Default Values](https://www.postgresql.org/docs/current/ddl-default.html)
- [SQLite Syntax: column-constraint](https://www.sqlite.org/syntax/column-constraint.html)

Currently, the parser only extracts the function name in function calls, leading to inappropriate conversions (https://github.com/k0kubun/sqldef/issues/388).

Hopefully, there are no mistakes in the implementation approach.


k0kubun (approved):
Thank you!

cockscomb:
Thank you for the quick review and release!

PR #433, opened 2023-08-31
Fix missing SET type members

cockscomb:
In the current version of sqldef, adding a column of SET type results in an ALTER TABLE statement with empty members. This PR fixes that issue.

### current

```sql
CREATE TABLE alarm (id BIGINT PRIMARY KEY);
```

### desired

```sql
CREATE TABLE alarm (
  id BIGINT PRIMARY KEY,
  dayOfWeek SET('Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun') NOT NULL
);
```

## Expected

```sql
ALTER TABLE `alarm` ADD COLUMN `dayOfWeek` set('Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun') NOT NULL AFTER `id`;
```

## Actural

```sql
ALTER TABLE `alarm` ADD COLUMN `dayOfWeek` set NOT NULL AFTER `id`;
```

### DIFF
--- a/cmd/mysqldef/tests.yml
+++ b/cmd/mysqldef/tests.yml
@@ -327,6 +327,16 @@ TableComment:
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COMMENT='都道府県マスター';
   output: |
     ALTER TABLE `prefecture` COMMENT = '都道府県マスター';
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
Thank you!

PR #438, opened 2023-09-03
Add Support for Creating and Deleting SQLite Triggers

shuymn:
Hello,

This PR adds support for creating and deleting SQLite Triggers. 
The specifications for each statement can be found here:

- [CREATE TRIGGER](https://www.sqlite.org/lang_createtrigger.html)
- [DROP TRIGGER](https://www.sqlite.org/lang_droptrigger.html)

To support these statements, the following changes have been made:

- Added functionality to retrieve the existing trigger configurations in SQLite Database, which was previously missing.
- Modified parser.y to be able to parse SQLite's CREATE TRIGGER statements.
- Ensured that the appropriate migration SQL is generated when triggers are created, modified or dropped.

For the generated CREATE TRIGGER statements, we used the input schema directly in SQLite, unlike MySQL and SQL Server where we generated SQL based on parsed results. This design decision was inspired by how CREATE TABLE statements are handled.

I'm not an expert in parser implementations, so if you have any suggestions for improvement, please feel free to point them out.

Thank you.


schema/generator.go
```diff
@@ -268,6 +268,18 @@ func (g *Generator) generateDDLs(desiredDDLs []DDL) ([]string, error) {
 		ddls = append(ddls, fmt.Sprintf("DROP EXTENSION %s", g.escapeSQLName(currentExtension.extension.Name)))
 	}
 
+	// Clean up obsoleted triggers
+	for _, currentTrigger := range g.currentTriggers {
+		if g.mode != GeneratorModeSQLite3 {
+			continue
+		}
```
shuymn, line 275: This PR is specific to SQLite, so I've left the status quo for databases other than SQLite. Specifically, those databases don't handle simply deleting a trigger instead of updating it. (Or is there something I'm missing?)

k0kubun (approved):
Thanks!

PR #440, opened 2023-09-07
Add support DATETIME fractional seconds part

cockscomb:
In the current version of sqldef, changing the column type from DATETIME to DATETIME(3) does not produce an ALTER TABLE statement. This PR fixes this comparison.

### current

```sql
CREATE TABLE users (
  id bigint(20) PRIMARY KEY,
  created_at datetime NOT NULL
);
```

### desired

```sql
CREATE TABLE users (
  id bigint(20) PRIMARY KEY,
  created_at datetime(3) NOT NULL
);
```

## Expected

```sql
ALTER TABLE `users` CHANGE COLUMN `created_at` `created_at` datetime(3) NOT NULL;
```

## Actural

```sql
```


k0kubun (approved):
Thank you!

Issue #455, opened 2023-10-27
psqldef panics on valid VIEW with nested CASE


chumaltd:
psqldef gets panic on parsing online VIEWs with nested CASE. 
This is split from #454, as they have different causes.

## Platform

* OS: Linux
* RDBMS: PostgreSQL
* Version: v0.16.9

### --export output

```bash
$ psqldef -Upostgres -hlocalhost -p5432 --export some_db

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xbf86e9]

goroutine 1 [running]:
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseExpr({{0x0?}, 0x0?}, 0x0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:293 +0x29
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseExpr({{0x0?}, 0x0?}, 0x0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:351 +0xbe6
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseResTarget({{0x203000?}, 0x0?}, 0xc0000c23c0)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:281 +0x23b
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseSelectStmt({{0x0?}, 0x0?}, 0xc000141b80)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:201 +0x1f6
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseViewStmt({{0x1?}, 0x68?}, 0xc000303f00)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:172 +0xaf
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseStmt({{0xc00040527f?}, 0x0?}, 0xc0003d0fa0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:76 +0x185
github.com/k0kubun/sqldef/database/postgres.PostgresParser.Parse({{0x2b?}, 0x40?}, {0xc000400000, 0x5498})
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:45 +0x195
[...]

Issue #454, opened 2023-10-23
psqldef panics, abort with syntax error on valid VIEWs


chumaltd:
Hi, I hit into 2 cases of psqldef parse error on VIEW.  
And, I extracted reproducible DDL for each one. They may look meaningless, but taken from actual use cases.

I think `--skip-view` option would be helpful for an escape hatch, as table operation looks much stable and Views can be `CREATE OR REPLACE`ed by psql.

## Platform

* OS: Linux
* RDBMS: PostgreSQL
* Version: v0.16.9

### --export output

```bash
$ psqldef -Upostgres -hlocalhost -p5432 --export some_db

2023/10/23 12:50:14 found syntax error when parsing DDL "CREATE VIEW public.any_array AS SELECT 1 AS "?column?" WHERE (1 = ANY (ARRAY[1, 4, 5]))": syntax error at position 79 near '1'
```

### Input SQL

```sql
CREATE VIEW any_array AS SELECT 1 WHERE 1 in (1, 4, 5);
```

PostgreSQL internally transforms this as follows:

```sql
CREATE VIEW any_array AS SELECT 1 WHERE 1 = ANY(ARRAY[1, 4, 5]);
```

### --export output

```bash
$ psqldef -Upostgres -hlocalhost -p5432 --export some_db

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xbf86e9]

goroutine 1 [running]:
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseExpr({{0x0?}, 0x0?}, 0x0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:293 +0x29
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseExpr({{0x0?}, 0x0?}, 0x0?)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:351 +0xbe6
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseResTarget({{0x203000?}, 0x0?}, 0xc0000c23c0)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:281 +0x23b
github.com/k0kubun/sqldef/database/postgres.PostgresParser.parseSelectStmt({{0x0?}, 0x0?}, 0xc000141b80)
        /home/runner/work/sqldef/sqldef/database/postgres/parser.go:201 +0x1f6
[...]

Issue #214, opened 2022-03-04
[mysqldef]feature request: ignore view option


tatsuo48:
Version: MySQL5.6.24

In my use case I want to apply one schema file to multiple schemas.
I want them all to have the same schema definition.

The exported schema file has a view, which contains the schema name.
export exmaple:
` mysqldef -h 0.0.0.0 -uroot -p schema1 --export > schema.sql`
```sql
CREATE TABLE `a` (
  `id` int(6) NOT NULL,
  `name` varchar(255) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;


CREATE VIEW a_view AS
SELECT
	`schema1`.`a`.`id` AS `id`,
	`schema1`.`a`.`name` AS `name`
FROM
	`schema1`.`a`;
```

If you apply this schema file to another schema, it will be considered a diff because the schema names are different.
apply other schemas exmaple:
` mysqldef -h 0.0.0.0 -uroot -p schema2  < schema.sql`
```sql
CREATE TABLE `a` (
  `id` int(6) NOT NULL,
  `name` varchar(255) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

CREATE VIEW a_view AS select `schema1`.`a`.`id` AS `id`,`schema1`.`a`.`name` AS `name` from `schema1`.`a`;
```

If possible, it would be nice if the schema name in the view was replaced as follows.
```sql
CREATE TABLE `a` (
  `id` int(6) NOT NULL,
  `name` varchar(255) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

CREATE VIEW a_view AS select `schema2`.`a`.`id` AS `id`,`schema2`.`a`.`name` AS `name` from `schema2`.`a`;
```

Or it would be nice to have the option to ignore the view when exporting or applying for such cases.
Would you consider adding an option?

k0kubun:
This sounds like a reasonable problem to be solved. Though I think the final resolution should not be an option but something like: if you do not specify database names in a view definition, sqldef will assume the abbreviated database name is the database you're connected to, as it should. You would still have to manually remove the database name in an exported schema to activate the feature, but I feel it'd be the most convenient interface.

[...]

PR #456, opened 2023-10-28
add --skip-view option to psqldef

chumaltd:
I just add `psqldef --skip-view` allong w/ #454 discussion, and confirmed this can skip parsing views including both #454 and #455 cases.

This patch mostly traces #214 implementation.
Trick or treat!


k0kubun (approved):
Thanks!

PR #460, opened 2023-11-14
[psqldef] Add --skip-extension option

hokaccha:
I have added the `--skip-extension` option to psqldef because there are cases where we don't want to manage extensions with sqldef.

@k0kubun Please review. If you approve, I will merge and release it myself.


k0kubun (approved):
:+1:

Issue #466, opened 2023-11-20
Add an ability to manage schema objects by schema restricted users.


akisssa:
In the case of managing tables by schema restricted user, we cannot create table objects in other schema. This is because users can see table metadata outside of current schema with `pg_catalog.pg_namespace`. So I want to add an ability to psqldef.
   
There may be some options to implement this.

1. Skip if not match the current schema (https://github.com/sqldef/sqldef/blob/master/database/postgres/database.go#L98) in tableNames func. But this option may affect users who manage objects in multiple schemas. 
```go
		if schema != d.GetDefaultSchema() {
			continue
		}
```
2. Add an option like `--targetSchema` to psqldef and use it with the example 1 .

## Platform

* OS: Linux 
* RDBMS:  PostgreSQL 
* Version: v0.16.12

## --export output

None

## prepare 

```sql
CREATE SCHEMA scha;
CREATE USER usra WITH ENCRYPTED PASSWORD 'usra';
ALTER ROLE usra SET search_path TO scha;
GRANT ALL ON SCHEMA scha to usra;

CREATE SCHEMA schb;
CREATE USER usrb WITH ENCRYPTED PASSWORD 'usrb';
ALTER ROLE usrb SET search_path TO schb;
GRANT ALL ON SCHEMA schb to usrb;
```

## Input SQL

schema_a

```
CREATE table tbl_a (id int);
```

schema_b

```
CREATE table tbl_b (id int);
```

schema_b

## Current output

schema_a

```
-- Apply --
CREATE table tbl_a (id int);
```

schema_b
```
Error on DumpDDLs: sql: Scan error on column index 0, name "column_name": converting NULL to string is unsupported
exit status 1
```

## Expected output

schema_a

```
-- Apply --
CREATE table tbl_a (id int);
```

schema_b

```
-- Apply --
CREATE table tbl_b (id int);
```

k0kubun:
(1) seems too specific to the one-schema-per-user model as you said. I'd prefer (2) since it seems like a generalization that can be achieved fairly easily.

As to the interface, WDYT about doing it over `--config` option? That configuration file already supports `target_tables`, so it seems more consistent to introduce something like `target_schema` there.

akisssa:
@k0kubun Supporting `target_schema` option on `--config` looks better for this use case. I will try to fix it.

akisssa:
Tried but it's scope is too large, so added a new `--targetSchema` option to the psqldef. Please check the pull request for details.

PR #467, opened 2023-11-21
[psqldef] Add --target-schema option

akisssa:
Ref. #466 

I have added `--target-schema` option in order to apply specific PostgreSQL `schema`'s tables.

@k0kubun Thanks for replying the issue. Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Application Options:
       --skip-view            Skip managing views/materialized views
       --skip-extension       Skip managing extensions
       --before-apply=        Execute the given string before applying the regular DDLs
+      --target-schema=       Apply DDLs only specific schema
       --config=              YAML file to specify: target_tables
       --help                 Show this help
       --version              Show this version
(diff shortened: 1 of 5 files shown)


k0kubun:
> https://github.com/sqldef/sqldef/issues/466
>  scope is too large

What do you mean by "scope is too large"? Isn't it possible to do the same thing but with reading values from a config file?

The idea behind `--config` was to prevent `psqldef` from having too many options that do not exist in `psql` as it tries to be as compatible as possible. Any option that does not exist in `psql` is heavily discouraged, and `--config` is an escape hatch to eventually manage sqldef-specific options in a single place.

akisssa:
> What do you mean by "scope is too large"? 

I thought the scope of changes would be vast. When I looked at it again, it wasn't like that.

> The idea behind --config was to prevent psqldef from having too many options that do not exist in psql as it tries to be as compatible as possible (to keep it easy to understand and use for new comers). 

I strongly positive to this. So I modify to use `--config`.

PR #500, opened 2024-03-28
Put the alter foreign key and index at the end of DDLs

hokaccha:
fix https://github.com/sqldef/sqldef/issues/405

@k0kubun The impact is significant, so I would appreciate your review.

### DIFF
--- a/cmd/psqldef/tests.yml
+++ b/cmd/psqldef/tests.yml
@@ -192,6 +192,29 @@ PosixRegexCheckConstraint:
     ALTER TABLE "public"."test" ADD CONSTRAINT test_posix_not_regex_check CHECK (posix_not_regex !~ '[a-z]');
     ALTER TABLE "public"."test" DROP CONSTRAINT test_posix_not_regex_ci_check;
     ALTER TABLE "public"."test" ADD CONSTRAINT test_posix_not_regex_ci_check CHECK (posix_not_regex_ci !~* '[a-z]');
+IndexAndForeignOrder:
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
makes sense to me.

hokaccha:
thanks!

yuku:
Notes for future developers. 

Using inline foreign key with independent unique constraint no longer works as expected:

```sql
CREATE TABLE dests (
  foo INT
);
ALTER TABLE dests ADD CONSTRAINT dests_unique UNIQUE (foo);

CREATE TABLE sources (
  foo INT REFERENCES dests (foo)
);
-- (output) there is no unique constraint matching given keys for referenced table "dests"
```

To address the issue, inline the unique constraint:

```sql
CREATE TABLE dests (
  foo INT,
  CONSTRAINT dests_unique UNIQUE (foo)
);

CREATE TABLE sources (
  foo INT REFERENCES dests (foo)
);
```

hokaccha:
@yuku Thank you. That's right. Or, you could do it like this.

```sql
CREATE TABLE dests (
  foo INT
);
ALTER TABLE dests ADD CONSTRAINT dests_unique UNIQUE (foo);

CREATE TABLE sources (
  foo INT
);
ALTER TABLE sources ADD CONSTRAINT sources_foo_fkey FOREIGN KEY (foo) REFERENCES dests (foo);
```

Either way, it would be better to standardize whether you define it inline or with an ALTER statement.

PR #501, opened 2024-03-29
Fix ALTER DDL to not add AFTER syntax when adding to last column.

tomtwinkle:
# Background
In the current sqldef, the AFTER syntax is always specified when adding a column other than the first.

```diff
CREATE TABLE users (
    id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
    name varchar(40) NOT NULL,
    created_at datetime NOT NULL,
+   updated_at datetime NOT NULL
);
```

```sql
ALTER TABLE `users` ADD COLUMN `updated_at` datetime NOT NULL AFTER `created_at`;
```

However, since the RDB behaves differently when FIRST/AFTER is specified or not specified, it may be desirable to execute without specifying AFTER.
Specifically, in MySQL, when adding a column in ALTER TABLE, the INSTANT Algorithm can be used when adding to the last column, but if BEFORE/AFTER is used, the INPLACE Algorithm is used (MySQL 8.0.29 or later ).

https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html
> The INSTANT algorithm can add a column at any position in the table. Before MySQL 8.0.29, the INSTANT algorithm could only add a column as the last column of the table.

There are cases where the INSTANT algorithm should be used instead of the INPLACE algorithm due to the specific bugs listed below.
https://docs.aws.amazon.com/AmazonRDS/latest/AuroraMySQLReleaseNotes/AuroraMySQL.Updates.3050.html
> Fixed an issue where the reader instance is unable to open a table, with ERROR 1146. This issue occurs when executing certain types of online Data Definition Language (DDL) while the INPLACE algorithm is being used on the writer instance.

# Solution

Fix to issue DDL without AFTER when adding to the last COLUMN of a table.

```diff
CREATE TABLE users (
    id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
    name varchar(40) NOT NULL,
    created_at datetime NOT NULL,
+   updated_at datetime NOT NULL
);
```

```sql
ALTER TABLE `users` ADD COLUMN `updated_at` datetime NOT NULL;
```

### DIFF
--- a/cmd/mysqldef/mysqldef_test.go
+++ b/cmd/mysqldef/mysqldef_test.go
@@ -217,6 +217,18 @@ func TestMysqldefAddColumnAfter(t *testing.T) {
 	)
 	assertApplyOutput(t, createTable, applyPrefix+"ALTER TABLE `users` ADD COLUMN `name` varchar(40) NOT NULL AFTER `id`;\n")
 	assertApplyOutput(t, createTable, nothingModified)
+
+	createTable = stripHeredoc(`
+		CREATE TABLE users (
+		  id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+		  name varchar(40) NOT NULL,
+		  created_at datetime NOT NULL,
+		  updated_at datetime NOT NULL
+		);
+		`,
+	)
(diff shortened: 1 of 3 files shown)


tomtwinkle:
Even if AFTER is specified, the last column will be INSTANT, so there is no need to specify it.

PR #512, opened 2024-04-26
Exclude a temporary table

chumaltd:
This PR excludes temporary tables from psqldef migration.

Having temporary tables on migration gets an ambiguous error:

```bash
2024/04/26 09:58:16 Error on DumpDDLs: sql: Scan error on column index 0, name "column_name": converting NULL to string is unsupported
```

This comes from temp tables not returning column defs.  
I think they should be excluded because we cannot say much about temporalities.

I just tried to add a test case, but couldn't find out a proper way to describe excluded DB objects. I confirmed it passes through existing tests.


k0kubun (approved):
Thanks!

PR #525, opened 2024-05-20
[psqldef] Fix the conditions for issuing REPLACE VIEW

hokaccha:
followup: https://github.com/sqldef/sqldef/pull/523

In the previous PR, it did not handle the case of adding a column in the middle, so I have corrected it.


cmd/psqldef/tests.yml
```diff
@@ -540,14 +540,93 @@ CreateViewCast:
...
+
+    CREATE VIEW v AS
+    SELECT
+      foo,
+      'x' as added,
+      bar
+     FROM hoge;
+  output: |
+    DROP VIEW "public"."v";
+    CREATE VIEW "public"."v" AS select foo, 'x' as added, bar from hoge;
```
hokaccha, line 618: This time we handled this test case.

PR #527, opened 2024-05-25
[mysqldef] Add Lock option

ethan-k:
Produce sql with lock option. 
Resolving https://github.com/sqldef/sqldef/issues/526 

for example, 

```
ALTER TABLE `users` ADD COLUMN `new_column` varchar(255) COLLATE utf8mb4_bin DEFAULT null AFTER `name`, LOCK=NONE;
```


k0kubun (approved):
:+1:

PR #544, opened 2024-07-17
[psqldef] Support '+' operator with intervals in column DEFAULT expressions

strager:


cmd/psqldef/tests.yml
```diff
@@ -984,6 +984,17 @@ RemoveDefaultExpression:
...
+  current: |
+    CREATE TABLE public.foo (
+      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '1 day'::interval)
+    )
+  desired: |
+    CREATE TABLE public.foo (
+      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '3 day'::interval)
+    )
+  output: |
+    ALTER TABLE "public"."foo" ALTER COLUMN "expires_at" SET DEFAULT current_timestamp + '3 day';
```
strager, line 997: I don't know why `::interval` is dropped. The ALTER TABLE command seems to work anyway. 🤷‍♀️

cmd/psqldef/tests.yml
```diff
@@ -984,6 +984,17 @@ RemoveDefaultExpression:
...
   output: |
     ALTER TABLE "public"."users" ALTER COLUMN "key" DROP DEFAULT;
+ChangeDefaultExpressionWithAddition:
+  current: |
+    CREATE TABLE public.foo (
+      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '1 day'::interval)
+    )
+  desired: |
+    CREATE TABLE public.foo (
+      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '3 day'::interval)
```
k0kubun, line 994: I wonder if this fixes the CI failure:

```suggestion
      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '1 day')
    )
  desired: |
    CREATE TABLE public.foo (
      expires_at timestamp with time zone NOT NULL DEFAULT (CURRENT_TIMESTAMP + '3 day')
```

cmd/psqldef/tests.yml
strager: I figured out why. Will upload a fixed version soon.

cmd/psqldef/tests.yml
strager: That doesn't work because PostgreSQL wants to re-add the ::interval cast.

I'll upload a fix soon.

k0kubun (approved):
:+1:

PR #545, opened 2024-07-18
[psqldef] Treat 'timestamptz' the same as 'timestamp with time zone'

strager:
When parsing a column with type 'timestamptz', we don't go through the code path which normalizes 'timestamptz' to 'timestamp with time zone'. This causes psqldef to try to change the type of a timestamptz column.

Use the normalization codepath by default, but include some compatibility checks to avoid changing existing tests.

### DIFF
--- a/database/postgres/parser.go
+++ b/database/postgres/parser.go
@@ -1071,40 +1071,48 @@ func (p PostgresParser) parseTypeName(node *pgquery.TypeName) (parser.ColumnType
 		}
 	}
 
-	if len(typeNames) == 1 {
-		columnType.Type = typeNames[0]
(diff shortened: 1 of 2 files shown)


database/postgres/parser.go
```diff
@@ -1071,40 +1071,48 @@ func (p PostgresParser) parseTypeName(node *pgquery.TypeName) (parser.ColumnType
 		}
 	}
 
-	if len(typeNames) == 1 {
```
strager, line 1074: This diff is a bit easier to read if you add `?w=1` to the end of the GitHub URL (or use `git diff -w` locally).

k0kubun (approved):

PR #502, opened 2024-03-29
[psqldef] Handle truncated auto generated constraint name correctly

yuku:
This PR enhances psqldef to accurately handle constraints without explicitly named column constraints. By default, Postgres generates constraint names based on the table and column names. However, due to the maximum constraint name length of 63 characters, names exceeding this limit are automatically truncated by Postgres. This update ensures psqldef properly manages these scenarios.

For a detailed illustration, please refer to the test case included in this patch.

PR #547, opened 2024-07-31
[psqldef] Handle truncated auto generated check constraint name correctly

yuku:
This pull request builds upon PR #502 by extending psqldef's capabilities to include support for **check constraints**. While PR #502 successfully introduced foreign key support, this subsequent PR aims to provide a more comprehensive solution for defining table schemas.


k0kubun (approved):
:+1:

PR #548, opened 2024-08-01
docker compose v2

johejo:
Fix https://github.com/sqldef/sqldef/actions/runs/10193267278/job/28197573316

```
line 1: docker-compose: command not found
```

See also

- https://github.com/actions/runner-images/issues/9692
- https://docs.docker.com/compose/compose-file/04-version-and-name/
- https://docs.docker.com/compose/compose-application-model/#the-compose-file


k0kubun (approved):
Thanks!

PR #555, opened 2024-09-03
[mysqldef] subtr and substring parser.y fix.

kzfk:
One part of the parser in “substr” was wrong.
> https://github.com/sqldef/sqldef/compare/master...kzfk52:sqldef:feature/substr_parser_fix?expand=1#diff-157e54ddd07b46a75c54cc2436fdc573d725f91c45e17c4fada5b39e74c86f32L3454

Also, although substr/substring can also use functions as its first argument, the parser has been changed to specify that functions can also be used, since the column names were specified.


k0kubun (approved):
Thank you!

PR #556, opened 2024-09-05
Add dump_concurrency option into config file

h6ah4i:
Hi, I noticed that `psqldef --export` (`psqldef`) takes about 2 minutes to run on my setup. The database is located in a distant region from my PC, and it contains 40 tables. This pull request introduces a new option that significantly reduces the time required for the database dump process.

---

This new ~~command-line~~ `dump_concurrency` option in the config file controls the concurrency level of queries during DB schema dump process.

It speeds up the DB schema dump process when enabled.

This option is particularly effective when the database is located remotely and contains many tables, as network latency becomes the dominant factor in the total duration of the dump process.

It is only effective for MySQL and PostgreSQL. This is because the MSSQL implementation generates table DDLs through string concatenation without executing many queries, and SQLite is an in-process database where its connections are (typically) not thread-safe.


h6ah4i:
Okay, I'll modify my code to make compatible with Go 1.19 👍

k0kubun:
Could we move this option to a filed in `--config`? We try to minimize the CLI's deviation from the original CLI (`mysql`, `psql`, `mssql`, `sqlite3`). To do so, newer options have been implemented as a field in the config file unless they exist in the original CLI as well.

If the original CLI has an option to specify this using a command-line flag or an environment variable, you may use the interface instead. Otherwise, we put such options in `--config` by default.

h6ah4i:
> We try to minimize the CLI's deviation from the original CLI (mysql, psql, mssql, sqlite3). To do so, newer options have been implemented as a field in the config file unless they exist in the original CLI as well.

Oh, I haven't noticed that policy! Moving this option to the config file sounds reasonable to me :+1:

How about moving this option to the config file as a dump_concurrency field, like the following?

```diff
var config struct {
	TargetTables    string `yaml:"target_tables"`
	SkipTables      string `yaml:"skip_tables"`
	TargetSchema    string `yaml:"target_schema"`
	Algorithm       string `yaml:"algorithm"`
	Lock            string `yaml:"lock"`
+	DumpConcurrency int    `yaml:"dump_concurrency"`
}
```

k0kubun:
That's what I meant :+1:

h6ah4i:
I've pushed a new commit and I tweaked the title/descriptions of this PR!

util/concurrent.go
```diff
@@ -0,0 +1,68 @@
+package util
```
k0kubun, line 1: I feel a bit uncomfortable introducing this package just for this purpose since people may start putting everything here going forward. Given that we use this only for database connections at the moment, how about moving this to `sqldef/sqldef/database` package (e.g. `database/concurrent.go`) instead?

k0kubun:
Can you write a test using `dump_concurrency` for psqldef and mysqldef?

util/concurrent.go
h6ah4i: I moved it to `database/concurrent.go` 👍

h6ah4i:
> Can you write a test using `dump_concurrency` for psqldef and mysqldef?

Tests were added in commit 641d2e6. Could you take a look at them?

tuzzmaniandevil:
Could a multi threaded approach like this also work for generating the DDLs?
Running `--dry-run` currently takes around 3 to 6 minutes for us on a good day. We do have a lot of tables though (250+)

k0kubun:
> Could a multi threaded approach like this also work for generating the DDLs?

I'm guessing the generator implementation is not a bottleneck for it. You might be able to check it by comparing the performance of `--export` and `--dry-run`. sqldef executes multiple database queries to dump the current schema, and I think it would be much slower than computing structs in the heap. It's worth testing this feature once released.

k0kubun (approved):

PR #560, opened 2024-09-20
[mysqldef] Fix ALGORITHM option being overwritten by LOCK in config

ysktir:
The previous implementation had a bug where specifying both ALGORITHM and LOCK in the configuration would result in only LOCK being applied. This PR fixes the issue to ensure both options are correctly used when specified.

```yaml
# config.yml
algorithm: inplace
lock: none
```

Before
```sql
ALTER TABLE `users` CHANGE COLUMN `name` `name` varchar(1000) DEFAULT NULL, LOCK=NONE;
```

After
```sql
ALTER TABLE `users` CHANGE COLUMN `name` `name` varchar(1000) DEFAULT NULL, ALGORITHM=INPLACE, LOCK=NONE;
```


ysktir:
It appears that the MSSQL crash on GitHub Actions' ubuntu-latest (20.04) is likely due to the issue described in the following links:
https://github.com/microsoft/mssql-docker/issues/868
https://github.com/microsoft/mssql-docker/issues/881

As suggested in the comments of these issues, I've confirmed that using either of the following images resolves the crash:
```
mcr.microsoft.com/mssql/server:2019-CU28-ubuntu-20.04
mcr.microsoft.com/mssql/server:2022-latest
```
Regarding the failing MySQL 8.0 CI, I believe this is unrelated to the MSSQL issue. The MySQL failure is likely caused by the test code being executed before the database becomes healthy. We may need to implement a proper health check or wait mechanism for MySQL as well.

k0kubun (approved):
:+1:

Issue #566, opened 2024-10-10
Out of memory in v0.17.20


komi1230:
## Platform

* OS: Linux
* RDBMS: PostgreSQL
* Version: v0.17.20

## Input SQL

Any schema file, even if empty schema file

## Current output

<details><summary>out of memory logs</summary>

```
fatal error: runtime: out of memory

runtime stack:
runtime.throw({0xb08fbc?, 0x203043?})
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/panic.go:1067 +0x48 fp=0x7ffeaffb2ab8 sp=0x7ffeaffb2a88 pc=0x46ea28
runtime.sysMapOS(0xc00dc00000, 0x100000000)
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/mem_linux.go:168 +0x11b fp=0x7ffeaffb2af8 sp=0x7ffeaffb2ab8 pc=0x41769b
runtime.sysMap(0xc00dc00000, 0x100000000, 0x1250788?)
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/mem.go:155 +0x34 fp=0x7ffeaffb2b18 sp=0x7ffeaffb2af8 pc=0x417074
runtime.(*mheap).grow(0x1240580, 0x80000?)
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/mheap.go:1539 +0x236 fp=0x7ffeaffb2b88 sp=0x7ffeaffb2b18 pc=0x42a1b6
runtime.(*mheap).allocSpan(0x1240580, 0x80000, 0x0, 0x1)
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/mheap.go:1244 +0x1b0 fp=0x7ffeaffb2c30 sp=0x7ffeaffb2b88 pc=0x429830
runtime.(*mheap).alloc.func1()
        /home/runner/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.2.linux-amd64/src/runtime/mheap.go:962 +0x5c fp=0x7ffeaffb2c78 sp=0x7ffeaffb2c30 pc=0x4292dc
runtime.systemstack(0x478a9f)
[...]

PR #567, opened 2024-10-11
[psqldef] Fixed out of memory error

fukasawah:
Related to Issue #566

With this update to go-pgquery, wazero will allocate larger memory during its initial operation, but by using a custom allocator, memory can now be allocated as needed.
Reference: https://github.com/wasilibs/go-pgquery/issues/31#issuecomment-2406412969


fukasawah:
Only psqldef 14 is failing... but I think it's temporary...

k0kubun (approved):
Thanks!

PR #573, opened 2024-11-16
mssqldef: Support UNIQUE CONSTRAINT

chi-bd:
Support UNIQUE CONSTRAINT both ALTER TABLE and COLUMN definition.
* BUGFIX: SQL Server's UNIQUE CONSTRAINT is also listed as INDEX, but it cannot delete by "DROP INDEX".
* NOTE: Avoiding syntax conflicts, "CLUSTER/NONCLUSTER" keyword is required for "ALTER TABLE ADD CONSTRAINT UNIQUE".


k0kubun (approved):
:+1:

PR #576, opened 2024-12-01
mssqldef: Support legacy SQL Server versions

chi-bd:
`sys.stats.is_incremental` only exists SQL Server 2014 (12.x) and above.
https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-stats-transact-sql?view=sql-server-ver16


k0kubun (approved):
Thanks!

PR #597, opened 2025-01-17
[psqldef] Fix altering partitions

### LINKED ISSUE #596: [psqldef] Partitions should be excluded
PostgreSQL partitioning has 2 table objects:

1. partitioned table, defining schema
2. partitions, having actual data

https://www.postgresql.org/docs/17/ddl-partitioning.html

PostgreSQL assumes that partitions' details are always derived from the partitioned table.  
So, psqldef operations on partitions should be simply excluded.

I'll attach the corresponding pull request soon.

This is related to #594 for better handling of partitioning functionality, but #594 is about partitioned tables.

## Background

sqldef assumes that migrations can be converted into several regular DDLs like `ALTER ...`  
When you modify partitions, you always need `DETACH`, `CREATE`, `INSERT`, then `ATTACH` combinations because data should be fit into the new configuration.  
I don't think such a situation is about sqldef way.

## Platform

OS: Linux
RDBMS: PostgreSQL
Version: v0.17.27

chumaltd:
This fixes #596. And I expect it mitigates #559 slowdown, too.

`relispartition` is descirbed:
https://www.postgresql.org/docs/current/catalog-pg-class.html

### DIFF
--- a/database/postgres/database.go
+++ b/database/postgres/database.go
@@ -93,6 +93,7 @@ func (d *PostgresDatabase) tableNames() ([]string, error) {
 		where n.nspname not in ('information_schema', 'pg_catalog')
 		and c.relkind in ('r', 'p')
 		and c.relpersistence in ('p', 'u')
+		and c.relispartition = false
 		and not exists (select * from pg_catalog.pg_depend d where c.oid = d.objid and d.deptype = 'e')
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
:+1:

PR #600, opened 2025-01-25
[psqldef] support exclusion constraints

hokaccha:
This is a continuation of the implementation from https://github.com/sqldef/sqldef/pull/528.

FYI @sanemat

fix https://github.com/sqldef/sqldef/issues/473
fix https://github.com/sqldef/sqldef/pull/528


sanemat:
Thanks 👍

PR #606, opened 2025-02-02
[psqldef] Handle case when schema is not specified in COMMENT ON clause

aki-s:
### About this PR

Partial fix for https://github.com/sqldef/sqldef/issues/605

This PR fixes only `Issue1: COMMENT ON clause is always seen as 'change' if no schema is specified.`.

### JFYI:

- `false positive` change is recorded at 
https://github.com/sqldef/sqldef/blob/912c9e249d3ccac4d936759a283d2c8e35a98704/schema/generator.go#L1778

- This PR  `modifies` `desired` (user specified) COMMENT clause if it didn't specify schema explicitly, so I felt this PR a bit strange approach although I've implemented. 
  - paserDDL https://github.com/sqldef/sqldef/blob/912c9e249d3ccac4d936759a283d2c8e35a98704/schema/generator.go#L68 should just parse, not modify user specified DDL. But current implementation already modifies using `defaultSchema`.
    - defaultSchema is introduced at https://github.com/sqldef/sqldef/pull/364/files#diff-d30ef67aecc70d57c316863ba09cbb56a864bc450da7db94a539dadcd81f4ee0R62
  - https://github.com/sqldef/sqldef/blob/912c9e249d3ccac4d936759a283d2c8e35a98704/schema/generator.go#L107-L108  should might have done comparison of parsedDDL somehow.

### DIFF
--- a/cmd/psqldef/tests.yml
+++ b/cmd/psqldef/tests.yml
@@ -570,6 +570,81 @@ CommentOnNonStandardDefaultSchema:
     COMMENT ON TABLE foo.users is 'users table updated';
     COMMENT ON COLUMN foo.users.id is 'users id';
   user: psqldef_user
+#; TODO: Enable to pass the following comment outed case 'CommentUnset'.
+#; 'Unsetting comment' is seen as 'change' false positively as of now (ref.
+#; https://github.com/sqldef/sqldef/issues/605 ), because the specified comment
+#; has no comment already before the application of sqldef,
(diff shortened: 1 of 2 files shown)


aki-s:
Issue 1 above is resolved by https://github.com/sqldef/sqldef/pull/606 .
I'll split Issue 2 to a newer ticket.

aki-s:
Thx

PR #613, opened 2025-02-21
Update go-pgquery

hokaccha:
Fix https://github.com/sqldef/sqldef/issues/611


hokaccha:
The following SQL failed to be parsed by the Postgres parser, which caused it to fall back to the generic parser.

```sql
CREATE TABLE users1 (
  c1 json default '[]'::json
);
```

PR #655, opened 2025-03-05
[psqldef] Recognize schema for view/materializedView

wreulicke:
[psqldef] Recognize schema for view/materializedView.

This fixes the bug that generates wrong alter statements for those when defined in a different schema.

<details>
<summary>outdated reproduction step</summary>
Creating a test in tests.yml is hard currently because it cannot set up schema outside of sqldef DDL.
"expected nothing is modified" error happens. 
And also, tests.yml doesn't support tests using a config.
So, I attached reproduction steps.

### Preparation

Create following three files and initialize database.

* init.sql
* config.yml
* schema.sql

init.sql
```
DROP DATABASE IF EXISTS sandbox;
CREATE DATABASE sandbox;
\c sandbox;
CREATE SCHEMA foo;

CREATE TABLE foo.users (
  id BIGINT PRIMARY KEY,
  name character varying(100)
);
CREATE TABLE foo.posts (
  id BIGINT PRIMARY KEY,
  name character varying(100),
  user_id BIGINT,
  is_deleted boolean
);
CREATE MATERIALIZED VIEW foo.view_user_posts AS SELECT p.id FROM (foo.posts as p JOIN foo.users as u ON ((p.user_id = u.id)));
```

config.yml
```
target_schema: |
  bar
```

schema.sql
```
CREATE SCHEMA bar;
CREATE TABLE bar.companies (
  id BIGINT PRIMARY KEY
);
```

initialize database.
```
psql -h localhost -U postgres < init.sql
```

## Expectation (After)

no alter statements about view/materializedView in different schema

```
$ go run ./cmd/psqldef --host localhost --user postgres --config config.yml sandbox < schema.sql
-- Apply --
CREATE SCHEMA bar;
CREATE TABLE bar.companies (
  id BIGINT PRIMARY KEY
);
```

## Actual (Before)

Unexpectedly, sqldef drops foo.view_user_posts.

```
$ go run ./cmd/psqldef --host localhost --user postgres --config config.yml sandbox < schema.sql
-- Apply --
CREATE SCHEMA bar;
CREATE TABLE bar.companies (
  id BIGINT PRIMARY KEY
);
DROP MATERIALIZED VIEW "foo"."view_user_posts";
```

</details>

### DIFF
--- a/cmd/psqldef/psqldef_test.go
+++ b/cmd/psqldef/psqldef_test.go
@@ -1489,6 +1489,43 @@ func TestPsqldefConfigIncludesTargetSchema(t *testing.T) {
 	assertEquals(t, apply, nothingModified)
 }
 
+func TestPsqldefConfigIncludesTargetSchemaWithViews(t *testing.T) {
+	resetTestDatabase()
+
+	mustExecuteSQL(`
+				CREATE SCHEMA foo;
+
+				CREATE TABLE foo.users (
+					id BIGINT PRIMARY KEY,
+					name character varying(100)
+				);
+				CREATE TABLE foo.posts (
+					id BIGINT PRIMARY KEY,
+					name character varying(100),
+					user_id BIGINT,
+					is_deleted boolean
+				);
+				CREATE VIEW foo.user_views AS SELECT id from foo.users;
+				CREATE MATERIALIZED VIEW foo.view_user_posts AS SELECT p.id FROM (foo.posts as p JOIN foo.users as u ON ((p.user_id = u.id)));
+    `)
+
+	schema := stripHeredoc(`
+				CREATE SCHEMA bar;
+				CREATE TABLE bar.companies (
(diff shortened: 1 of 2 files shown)


k0kubun:
> Creating a test in tests.yml is hard currently because it cannot set up schema outside of sqldef DDL.
> "expected nothing is modified" error happens.
> And also, tests.yml doesn't support tests using a config.

As an escape hatch that needs configs for testing, you may write such kinds of tests in `psqldef_test.go`. You may put such a test under the "following tests are for CLI" comment (because `--config` is a CLI option). Tests above that comment are ones that should be migrated to tests.yml, so it'd be nice to add that after that comment.

wreulicke:
> psqldef_test

@k0kubun 
I see. I will try it.

k0kubun (approved):
Thanks!

PR #668, opened 2025-04-23
Add skip_views option in config file

suzuito:
Please review the changes, and let me know if you have any feedback or suggestions. Thank you!

Thanks!

## Add `skip_views` option in config file

This pull request introduces a new feature to allow skipping views and materialized views during schema management. The changes include updates to configuration handling, schema generation, and testing to support the new `skip_views` functionality.

`skip_views` is available in config YAML file. This option is supported for PostgreSQL only.
(It actually works with MySQL and sqlite too, but I haven't tested it so it may behave unexpectedly.)
The reason why MySQL and sqlite are not supported is because I don't have knowledge of SQL for these databases, and because the tasks I am working on do not use MySQL and sqlite.

## Fix CI error

This pull request updates the GitHub Actions workflow configuration to use a newer version of the Ubuntu runner.

Because https://github.com/sqldef/sqldef/actions/runs/14631630554/job/41054848619

> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/sqldef.yml
+++ b/.github/workflows/sqldef.yml
@@ -12,7 +12,7 @@ on:
       - reopened
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         target:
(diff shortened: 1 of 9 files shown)


k0kubun (approved):
Thanks!

PR #695, opened 2025-06-20
Add `/v2` suffix to module name for Go Modules v2 compatibility

utah-KT:
This PR updates the module name to include the `/v2` suffix, following Go Modules versioning conventions for major version 2.

- Changed the `go.mod` module path to `github.com/sqldef/sqldef/v2`
- Updated all import paths to use `/v2`
- No functional changes; this is to ensure correct versioning and import behavior for v2 and above

Please update your import paths accordingly when upgrading to v2.


k0kubun (approved):
:ok_hand:

PR #705, opened 2025-07-16
psqldef: Emit Foreign Key Constraints Last to Support CREATE TABLE Style (Follow‑up to #500)

yuku:
This PR updates the DDL generation so that **`ALTER TABLE … ADD CONSTRAINT … FOREIGN KEY` statements are emitted last**, after all other schema changes.  
It adds a new test case (`ForeignKeyConstraintsAreEmittedLast`) to confirm this behavior.

### Background & Motivation

We maintain a very large schema (thousands of lines) in Git and continuously run `psqldef` to migrate our databases.  
Our style is to **write foreign keys inline in `CREATE TABLE`**, rather than as separate `ALTER TABLE` statements.

`psqldef` already supports the style of emitting standalone `ALTER TABLE ADD CONSTRAINT FOREIGN KEY` statements, as introduced in https://github.com/sqldef/sqldef/pull/500.  
This PR is a follow-up that makes the `CREATE TABLE` style easier to use in large, split schemas.

When splitting a large schema into multiple files, it is hard to guarantee that concatenated files are ordered so that all referenced tables exist before their constraints are applied.  
Currently, foreign key `ALTER TABLE` statements are interleaved with other DDLs, which complicates zero‑from‑scratch migrations.

### What This PR Enables

With this change, `psqldef` now delays foreign key constraints until after all other DDLs have been emitted.  
This enables the following workflow:

1. Create all tables first (even with temporary dummy columns).
2. Apply the merged schema generated from arbitrarily ordered files.
3. `psqldef` will:
   * Add new columns, primary keys and indexes,
   * Drop dummy columns,
   * **Finally add foreign key constraints at the very end.**

### QA

🔴 → 🟢 

image

### DIFF
--- a/cmd/psqldef/tests.yml
+++ b/cmd/psqldef/tests.yml
@@ -297,6 +297,27 @@ IndexAndForeignOrder:
     );
     CREATE UNIQUE INDEX uniq_idx ON users (id);
     ALTER TABLE posts ADD CONSTRAINT "user_id_key" FOREIGN KEY ("user_id") REFERENCES users ("id");
+ForeignKeyConstraintsAreEmittedLast:
+  current: |
+    CREATE TABLE foos ( dummy_column text );
+    CREATE TABLE bars ( dummy_column text );
+  desired: |
+    CREATE TABLE foos (
+      foo_id bigint PRIMARY KEY,
+      bar_id bigint NOT NULL REFERENCES bars (bar_id)
+    );
+    CREATE TABLE bars (
+      bar_id bigint PRIMARY KEY
+    );
+  output: |
+    ALTER TABLE "public"."foos" ADD COLUMN "foo_id" bigint NOT NULL;
+    ALTER TABLE "public"."foos" ADD COLUMN "bar_id" bigint NOT NULL;
+    ALTER TABLE "public"."foos" ADD PRIMARY KEY ("foo_id");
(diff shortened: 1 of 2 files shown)


yuku:
@hokaccha (as the author of #500) could you review this change?

cmd/psqldef/tests.yml
```diff
@@ -297,6 +297,27 @@ IndexAndForeignOrder:
...
+    CREATE TABLE bars (
+      bar_id bigint PRIMARY KEY
+    );
+  output: |
+    ALTER TABLE "public"."foos" ADD COLUMN "foo_id" bigint NOT NULL;
+    ALTER TABLE "public"."foos" ADD COLUMN "bar_id" bigint NOT NULL;
+    ALTER TABLE "public"."foos" ADD PRIMARY KEY ("foo_id");
+    ALTER TABLE "public"."bars" ADD COLUMN "bar_id" bigint NOT NULL;
+    ALTER TABLE "public"."bars" ADD PRIMARY KEY ("bar_id");
+    ALTER TABLE "public"."foos" ADD CONSTRAINT "foos_bar_id_fkey" FOREIGN KEY ("bar_id") REFERENCES "public"."bars" ("bar_id");
```
yuku, line 318: In `master`, the `ADD FOREIGN KEY` occurs before adding the destination column so it fails.

> image
>
> https://github.com/sqldef/sqldef/actions/runs/16308029916/job/46058064328

PR #727, opened 2025-09-06
Add column rename support with `-- @rename from=<old_name>` annotation

gfx:
This PR introduces a new annotation syntax `-- @rename from=<old_name>` (where `<old_name>` is a bare identifier or double-quoted one) to explicitly handle column renames in schema migrations. Previously, column renames were treated as drop+add operations, which could lead to unexpected data loss.

Key features:

- Supports all the SQL engines: MySQL, PostgreSQL, SQLite3, and MSSQL
- Handles columns with special characters, spaces, and dots
- Validates against conflicting renames (prevents renaming to existing columns)
- Idempotent - no-op when column already has the target name
- Allows type changes alongside renames
  - Because SQLite3 does not support to change the type of a column, sqlite3def generates a bit more complicated "add-update-drop" queries. See the added tests for SQLite3.

Example usage:

```sql
CREATE TABLE users (
  id bigint NOT NULL,
  user_name text, -- @rename from=username
  age integer
);
```

This generates appropriate rename DDL statements instead of drop/add, preserving data during migrations. See the test data for details.


k0kubun (approved):
Interesting. Thank you for proposing this feature. Let's try this.

PR #731, opened 2025-09-07
Add table rename support with `@rename from=<old_table_name>` magic comments

gfx:
With the similar syntax as https://github.com/sqldef/sqldef/pull/727

For example,

```sql
CREATE TABLE users ( -- @rename from=user_accounts
  id bigint NOT NULL,
  username text,
  age integer
);
```

will create:

```sql
ALTER TABLE `user_accounts` RENAME TO `users`;
```

See for README.md and the tests for details.


k0kubun (approved):
:+1:

PR #737, opened 2025-09-11
psqldef: support IN and UNIQUE constraints

gfx:
Close https://github.com/sqldef/sqldef/discussions/640


gfx:
@k0kubun 

Sure. Will do.

Next, I'll tweak the release workflow to keep immutable releases turned on.

PR #781, opened 2025-09-25
allow update statements without from clause inside triggers

osjupiter:
### Description
This fixes https://github.com/sqldef/sqldef/discussions/780

When parsing triggers containing UPDATE statements without FROM clauses, the parser was incorrectly adding `FROM dual` to the UPDATE statement, which is not valid MySQL syntax and would cause execution errors.

### Changes
- Modified the parser to handle empty FROM clauses properly by:
  - Changing `from_opt` rule to return empty `TableExprs{}` instead of automatically adding `dual` table
  - Updated `Update.Format()` to only include FROM clause when `TableExprs` is not empty using new `isEmpty()` method
  - Updated `Select.Format()` to add `FROM dual` only when FROM clause is actually empty (preserving existing SELECT behavior)


k0kubun (approved):
Thanks!

PR #847, opened 2025-10-02
psqldef: CREATE TABLE ... LIKE ...

gfx:
Resolve https://github.com/sqldef/sqldef/discussions/632


gfx:
Will give it a try with different ways.

PR #1131, opened 2026-02-09
Add MySQL CREATE EVENT / ALTER EVENT / DROP EVENT support

Schnouki:
Hey folks. While trying to use sqldef in an existing project, I was blocked by the fact that one of our MySQL databases uses `CREATE EVENT`, which wasn't supported. So, having no context on the sqldef codebase, I used Claude Code to add support for these events.

---

This PR adds support for MySQL scheduled events to mysqldef, following the same patterns used for triggers and views.

### What this does

- **Parse** `CREATE EVENT` statements including `DEFINER`, `ON SCHEDULE` (both `EVERY` and `AT`), `ON COMPLETION [NOT] PRESERVE`, `ENABLE`/`DISABLE`, `COMMENT`, and `BEGIN...END` compound bodies
- **Generate idempotent DDLs**: create new events, alter existing ones (via `ALTER EVENT`, not DROP+CREATE), and drop removed events
- **Export** events from a live MySQL database via `SHOW CREATE EVENT`
- **Normalize** MySQL defaults for comparison: MySQL's `SHOW CREATE EVENT` always adds `DEFINER`, `STARTS`, `ON COMPLETION NOT PRESERVE`, and `ENABLE` even when not originally specified. The comparison logic handles this so that events without these clauses are correctly recognized as matching.

### Design decisions

- **`ALTER EVENT` for modifications** rather than `DROP + CREATE`, which avoids losing the event's execution history and is more consistent with how MySQL handles events. `ALTER EVENT` always emits all clauses (ON COMPLETION, ENABLE/DISABLE, COMMENT) with MySQL defaults for empty values, because `ALTER EVENT` only modifies specified clauses and preserves the rest.
- **Schedule stored as raw string** instead of deeply parsed AST. This avoids grammar complexity for `AT timestamp + INTERVAL ... + INTERVAL ...` chains while still supporting all schedule forms. Compound interval units (`DAY_HOUR`, `YEAR_MONTH`, etc.) are handled via `table_id` which already accepts `UNUSED` tokens -- no new token declarations needed.
- **Reuses `trigger_statements`** grammar for the event body, since events use the same `BEGIN...END` block syntax.
- **DEFINER is parsed but excluded from comparison**, since MySQL normalizes it in `SHOW CREATE EVENT` output and the user-specified definer may differ from what MySQL stores.
- **`event_definer_user` grammar rule** handles both single-quoted (`'user'@'host'` from user SQL) and backtick-quoted (`` `user`@`host` `` from `SHOW CREATE EVENT`) definer formats, plus `CURRENT_USER`. This is needed because the tokenizer treats `@` as an identifier character, not a literal token.

### Files changed

| File | Change |
|------|--------|
| `parser/token.go` | 13 new keywords (event, schedule, every, at, do, ...) |
| `parser/parser.y` | Grammar rules for `CREATE EVENT` and helper rules |
| `parser/node.go` | `CreateEvent` action and `Event` AST struct |
| `schema/ast.go` | Schema-level `Event` type |
| `schema/parser.go` | `CreateEvent` → schema `Event` conversion |
| `schema/generator.go` | Event aggregation, comparison, `ALTER EVENT` generation, `DROP EVENT`, normalization helpers |
| `database/mysql/database.go` | `events()` export via `SHOW CREATE EVENT`, wired into `ExportDDLs` |
| `cmd/mysqldef/tests_events.yml` | 12 test cases (new file) |

### Parser conflicts

4 new shift/reduce conflicts (16 → 20), all from reusing `trigger_statements` for the event body -- same pattern as the existing trigger rules.

### Test plan

- [x] YAML test suite against real MySQL -- 12 tests:
  - Create: simple with COMMENT, DEFINER + ON COMPLETION PRESERVE, compound body (`BEGIN...END`), compound interval (`DAY_HOUR`), AT schedule
  - Alter: schedule, body, ON COMPLETION, COMMENT, DISABLE, multiple changes at once
  - Drop
- [x] Full mysqldef test suite -- no regressions
- [x] sqlite3def, parser, schema unit tests -- no regressions
- [x] `make build`, `make format`, `make lint` -- all clean
- [ ] TiDB: all event tests use `flavor: "!tidb"` (TiDB does not support events)
- [ ] MariaDB: untested locally, should work (MariaDB supports events with the same syntax)


gfx:
Can you merge master and regen the parser?

Schnouki:
@gfx Apologies for the delay; merged master and regenerated the parser. Thanks!

gfx:
@Schnouki Sorry! Conflict again. Can you resolve it once more?

Schnouki:
@gfx Done, thanks!

modasserbillah:
Hey @gfx the branch is updated again and the CI is green, can we get this in please? It would help us use `sqldef` much smoothly, thanks!

gfx:
Merging!

PR #1191, opened 2026-05-01
[FIX] PG: Unflatten view export

FinnLang-Kettner:
Postgesql views are normally completely flattened (1 long line) during export.
By removing the string manipulation from the view export we get the unaltered view exported from Postegresql.


gfx:
Absolutely LGTM.

Can you make CI green?

FinnLang-Kettner:
Yes, it seems I was a bit eager on the delete trigger.

PR #1192, opened 2026-05-01
feat: support comments on schema

FinnLang-Kettner:
I've added support for comments on schemas. However I'm not sure if we need more tests or if I have gotten the naming correct (e.g. is "OBJECT_SCHEMA" the appropriate ObjectType?).


FinnLang-Kettner:
I'm not sure what the problem is with parser.go. If I try to regenerate it from parser.y in the PR I get the same result as is in the PR.

FinnLang-Kettner:
Could someone please try to generate parser/parser.go from parser/parser.y? I have tried it from Windows and wsl with different results(?) but none of them match what is expected from what github makes.

PR #1201, opened 2026-05-03
feat(MySQL): add --bulk-alter to fuse ALTER TABLE actions per table 

exoego:
# motivation
To reduce DDL locks on large tables for every `ALTER TABLE` statement, multiple actions are fused into a single `ALTER TABLE t act1, act2, ...`.

# note

- <del>PostgreSQL and</del> MySQL only. PostgreSQL already uses a transaction for DDL. SQLite3 and MSSQL grammar seemingly disallow comma-separated action lists, so the flag is a documented no-op.
- CREATE/DROP INDEX and ADD/DROP FOREIGN KEY are unchanged; only the per-table inter DDLs and cleanup-phase ALTER TABLE statements are fused.
- RENAME forms (RENAME COLUMN / RENAME TO / RENAME CONSTRAINT) cannot be combined per PostgreSQL/MySQL grammar and remain standalone.

### DIFF
--- a/cmd-mysqldef.md
+++ b/cmd-mysqldef.md
@@ -26,6 +26,7 @@ Application Options:
       --export                      Just dump the current schema to stdout
       --enable-drop                 Enable destructive changes such as DROP for TABLE, SCHEMA, ROLE, USER, FUNCTION, PROCEDURE, TRIGGER, VIEW, INDEX, SEQUENCE, TYPE
       --skip-view                   Skip managing views (temporary feature, to be removed later)
+      --bulk-alter                  Bundle multiple ALTER TABLE actions on the same table into a single statement
       --before-apply=SQL            Execute the given string before applying the regular DDLs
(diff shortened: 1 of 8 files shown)


gfx:
Thanks for the PR!

Two requests before this can go in:

1. Please drop this from psqldef and keep it MySQL-only. The motivation is reducing DDL locks, but in PostgreSQL the first `ALTER TABLE` in a transaction already holds `ACCESS EXCLUSIVE` until COMMIT, so merging statements does not shorten the lock window. psqldef runs DDL in a transaction by default, so Pg users should rely on that instead. (And for `--disable-ddl-transaction` cases like Aurora DSQL, turning many short locks into one long lock is not clearly a win either.)

2. Please merge at the AST level instead of re-parsing the generated query texts. This project already includes a DDL parser, as you know, so the ALTER actions should be combined where they are emitted rather than by scanning the output strings.

exoego:
@gfx sorry for the late response. Updated PR so.

PR #1214, opened 2026-05-16
feat: add --check flag for CI gating of schema drift

TiBata:
## Summary

`--dry-run` always exits 0 whether the schema is in sync or not, so it cannot be used directly as a CI gate. There is no existing flag that combines "print the diff" with "exit non-zero on diff" — `--export` dumps current schema only, and `--apply` actually mutates the database.

This PR adds a `--check` flag (across all four commands: `mysqldef`, `psqldef`, `mssqldef`, `sqlite3def`) that:
- Behaves like `--dry-run` (uses the dry-run database wrapper, prints the DDL header + statements, makes no changes to the database).
- Exits with code **2** when DDL would be applied, **0** when the schema is already in sync.

Exit code **2** follows the convention of `diff(1)` and `terraform plan -detailed-exitcode`:
- `0` = no changes
- `2` = changes detected
- `1` stays reserved for actual tool errors (existing behavior)

This lets CI scripts distinguish "schema needs update" from "tool failed":

```bash
mysqldef --check db < schema.sql
case $? in
  0) echo "schema in sync" ;;
  2) echo "schema drift detected, see diff above"; exit 1 ;;
  *) echo "mysqldef error"; exit 1 ;;
esac
```

The exit code is exposed as `sqldef.CheckExitCode` for downstream Go consumers and documented in each command's `cmd-*.md`.

## Implementation

~10 lines in `sqldef.go`:
- `Options.Check bool` field
- `CheckExitCode` constant (= 2)
- After `RunDDLs` returns, if `options.Check`, call `os.Exit(CheckExitCode)`

Plus the flag registration in each of the 4 command frontends. The dry-run database plumbing is reused verbatim — no separate code path. No new tests infrastructure.

## Test coverage

`TestMysqldefCheck` in `cmd/mysqldef/mysqldef_test.go` covers the three observable behaviors:
1. Schema in sync → exit 0, `-- Nothing is modified --`
2. Schema differs → exit 2 (`sqldef.CheckExitCode`), DDL printed under `-- dry run --` header
3. Database unchanged after `--check` (verified by re-running `--check` and asserting it still reports the same diff — proving the first call didn't apply anything)

The core logic lives in shared `sqldef.Run()`, so the integration test for one command exercises the same path as the other three.

## Verified

- Full `TestApply` + `TestMysqldef*` suites: **0 failures** on MySQL 8.4 and only the 4 pre-existing `*VectorIndex*` baseline failures on MariaDB 10.11 (unrelated — MariaDB 10.11 doesn't support `VECTOR`).
- Manual smoke test against MariaDB confirms exit code 0 / 2 split + DB stays untouched.

## Scope note

This PR is intentionally minimal — just the flag, the exit code, the docs, and a test. No mutual-exclusion enforcement between `--check` / `--apply` / `--export` (the existing flags don't enforce that either; `--apply` simply takes precedence in the current code path, and `--check` slots into the same branch as `--dry-run`).

### DIFF
--- a/cmd-mssqldef.md
+++ b/cmd-mssqldef.md
@@ -18,6 +18,7 @@ Application Options:
       --password-prompt       Force MSSQL user password prompt
       --file=FILENAME         Read desired SQL from the file, rather than stdin (default: -)
       --dry-run               Don't run DDLs but just show them
+      --check                 Like --dry-run, but exit with code 2 when DDL would be applied (useful as a CI gate to detect schema drift)
       --apply                 Apply DDLs to the database (default, but will require this flag in future versions)
       --export                Just dump the current schema to stdout
       --enable-drop           Enable destructive changes such as DROP for TABLE, SCHEMA, ROLE, USER, FUNCTION, PROCEDURE, TRIGGER, VIEW, INDEX, SEQUENCE, TYPE
--- a/cmd-mysqldef.md
+++ b/cmd-mysqldef.md
@@ -21,6 +21,7 @@ Application Options:
       --enable-cleartext-plugin     Enable/disable the clear text authentication plugin
       --file=FILENAME               Read desired SQL from the file, rather than stdin (default: -)
       --dry-run                     Don't run DDLs but just show them
+      --check                       Like --dry-run, but exit with code 2 when DDL would be applied (useful as a CI gate to detect schema drift)
       --apply                       Apply DDLs to the database (default, but will require this flag in future versions)
       --export                      Just dump the current schema to stdout
       --enable-drop                 Enable destructive changes such as DROP for TABLE, SCHEMA, ROLE, USER, FUNCTION, PROCEDURE, TRIGGER, VIEW, INDEX, SEQUENCE, TYPE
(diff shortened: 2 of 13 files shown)


TiBata:
## On the codecov/patch failure

The patch coverage shows 18.18% because:

1. The CLI flag-wiring code in `cmd/*/.go` is only exercised when the compiled binary runs as a subprocess via `tu.Execute`. Go's coverage tooling doesn't capture coverage in child processes, only in the parent test process. This is why all `cmd/*/.go` files sit at 0% baseline coverage in this repo today.
2. The new test files (`TestMysqldefCheck` / `TestPsqldefCheck` / `TestMssqldefCheck` / `TestSQLite3defCheck`) are themselves excluded from coverage by `.codecov.yml`'s `**/*_test.go` ignore rule.

All four frontends have a dedicated `Test*Check` that asserts the three observable behaviors — exit 0 when in sync, exit 2 (== `sqldef.CheckExitCode`) when DDL would be applied, and database untouched after `--check` (proven by re-running `--check` and asserting the same diff still appears).

The behavior is verified end-to-end through every supported flavor; the coverage tool just can't see across the subprocess boundary. Happy to refactor toward in-process testing if that's preferred for future work, but it's a project-wide concern rather than something specific to this PR.

PR #1217, opened 2026-05-18
fix: extension is a non-reserved keyword

infogulch:
Fixes #1206

See:
* Postgres docs on sql keywords (non-reserved keyword): https://www.postgresql.org/docs/current/sql-keywords-appendix.html#:~:text=EXTENSION
* MySQL (not mentioned): https://dev.mysql.com/doc/refman/5.7/en/keywords.html
* SQLite (not mentioned): https://sqlite.org/lang_keywords.html

Created with separate test/fix commits to show red/green workflow.


infogulch:
Both this PR and #1216 are branched from master. This might be ok or even preferred for some repos, but here I suspect there may be a conflict in parser.go or a failing integrity check.

So I will pause here and wait for feedback on the PRs and direction on how to resolve integrity check failures (which may just be "rerun `make parser` and push").

I love declarative sql, thank you for making sqldef!

PR #1224, opened 2026-05-27
parser: fix Format silently dropping constraints in CREATE TABLE round-trip

178inaba:
## Why

`parser.String(*parser.DDL)` (CREATE TABLE round-trip) silently drops several
kinds of constraints that the AST actually carries:

- **Table-level** `CHECK` / `FOREIGN KEY` / `EXCLUDE` — `TableSpec.Format`
  only iterated `Columns` and `Indexes`. The three sub-element types
  (`CheckDefinition`, `ForeignKeyDefinition`, `ExclusionDefinition`) had no
  `Format` methods at all.
- **Column-level** inline `REFERENCES` — `ColumnType.Format` never emitted
  any of the 7 `Reference*` fields, so
  `a int REFERENCES u(id) ON DELETE CASCADE MATCH FULL DEFERRABLE INITIALLY DEFERRED`
  collapsed to `a int`.

While addressing those, three related parser bugs surfaced and are fixed
together because each is on the round-trip path:

- `EXCLUDE ... WHERE (pred)` parsed the parens into a `ParenExpr`, causing
  `WHERE ((...))` (double parens) on emission.
- `FOREIGN KEY ... MATCH FULL` was a syntax error: `match_type_opt` existed
  in the grammar but was not wired into table-level `FOREIGN KEY` rules.
- `deferrable_opt` / `initially_deferred_opt` returned `BoolVal`, collapsing
  three grammar states (absent, affirmative, negative) into two. This made
  it impossible to emit inline REFERENCES correctly: a naive
  `if pointer != nil` guard would either drop user-written `NOT DEFERRABLE`
  or attach `not deferrable initially immediate` to every plain
  `REFERENCES u (cols)` input.

## What

- `TableSpec.Format` now emits `Checks` / `ForeignKeys` / `Exclusions` after
  `Columns` / `Indexes` with the existing `,\n\t<value>` layout.
- New `Format` methods on `*CheckDefinition`, `*ForeignKeyDefinition`,
  `*ExclusionDefinition` (all lowercase keywords to match sibling style).
  A shared `formatFKColumnList` helper supports PostgreSQL 18+
  `PERIOD <col>` on the last column.
- New `exclude_where_opt` grammar rule that consumes the parens
  syntactically, removing the prior `ParenExpr` workaround in Format.
- `match_type_opt` threaded into the four `foreign_key_without_options`
  alternatives so `ForeignKeyDefinition.Match` is populated.
- `deferrable_opt` / `initially_deferred_opt` / `deferrable_clause` /
  `initially_clause` now return `*BoolVal`, with `nil` for the empty
  alternative. Column-level inline REFERENCES Alt 1 assigns the pointer
  directly; table-level UNIQUE call sites keep their plain-`bool`
  `ConstraintOptions` field via nil-guarded dereference, so existing
  UNIQUE / `fk_defer_opts` behavior is unchanged.
- `ColumnType.Format` emits inline REFERENCES between the `Check` and
  `KeyOpt` blocks, mirroring `ForeignKeyDefinition.Format` and honoring
  the tri-state `*BoolVal` pointers.
- 27 round-trip cases in `TestTableSpecRoundTripConstraints` plus
  `wantContains` / `wantNotContains` assertions (idempotency alone
  doesn't catch "Format drops everything," which still round-trips
  stably). 5 grammar pin cases in `TestInlineReferencesDeferrableGrammar`
  for the tri-state contract. Parser conflict count tracks the master
  baseline (21 shift/reduce, 505 reduce/reduce) — this PR adds no new
  conflicts. (The reduce/reduce baseline rose from 504 to 505 on master
  via #1219, which is independent of this PR.)

## Why this hasn't broken production

`schema/` and `database/<engine>/` traverse the AST nodes directly
(`column.references`, `tableSpec.Checks`, etc.) and never call
`parser.String(*DDL)`. A grep for production callers turns up only
`schema/parser.go:237` (`Domain.DataType`, which does not populate
`Reference*` fields) and `schema/generator.go:4888`
(`CastExpr.Type`, which does not populate the affected fields either).
Constraint diffing and DDL generation across the four sqldef commands
are therefore unaffected by this PR — the change is to AST-level
round-trip correctness, which the parser package implicitly promises.

## Design notes

- `Match` / `OnDelete` / `OnUpdate` are emitted via `.Name` string
  literals rather than `%v` because their values (`"MATCH FULL"`,
  `"SET NULL"`, `"NO ACTION"`) contain spaces, and `Ident.Format`
  would back-quote them.
- Lossy round-trip in `ColumnType` (Array / DisplayWidth / Srid /
  AutoRandom / Generated / Identity) is **out of scope**; tracked
  separately. Inline REFERENCES is included here because the grammar
  fix it requires is symmetric to the table-level FK fix that drove
  this branch.
- Table-level `ConstraintOptions.Deferrable bool` /
  `InitiallyDeferred bool` still collapse three states into two for
  UNIQUE / PRIMARY KEY / `fk_defer_opts` paths. Fixing that requires
  changing struct field types to `*bool` with `schema/` consumer
  audit, which is large enough to warrant a separate PR. The follow-
  up is drafted locally.

## Test plan

- [x] `go test ./parser/...` passes (verified locally)
- [x] `go test ./schema/...` passes (verified locally)
- [x] `go test ./cmd/sqlite3def/` passes (verified locally)
- [x] `make build` / `make format` / `make lint` pass (verified locally)
- [x] `make parser-v` reports conflict counts matching master baseline:
      21 shift/reduce, 505 reduce/reduce (verified locally after rebase
      onto master #1219)
- [x] CI verifies `make test-all-flavors` (MySQL / PostgreSQL / MSSQL /
      MariaDB) — requires live DBs, deferred to CI


178inaba:
Heads-up on this 0% report: it reflects a gap in CI's coverage upload, not a gap in this PR's tests.

The matrix here only runs the engine-rooted suites (`make test-{mysqldef,psqldef,sqlite3def,mssqldef}`), each of which executes `./cmd/<engine>` (+ `./database/<engine>`) and nothing else. `./parser`'s own tests are never executed in CI — `make test-core` (which runs `./parser ./schema ./util`) exists in the Makefile but isn't wired into `.github/workflows/sqldef.yml`. So the new `Format` methods added in this PR are exercised by `TestTableSpecRoundTripConstraints` (27 cases) and `TestInlineReferencesDeferrableGrammar` (5 cases) in `parser/parser_test.go`, but the resulting coverage profile is simply never uploaded.

Running `go test -coverpkg=./parser ./parser` locally shows the new methods at 70–90% line coverage (e.g. `TableSpec.Format` 84.6%, `*ForeignKeyDefinition.Format` 92.3%, `*CheckDefinition.Format` 85.7%, `formatFKColumnList` 87.5%).

Compounding this on the cmd side: as described in the PR body under "Why this hasn't broken production," the `parser.String(*DDL)` path is not reached by any production caller — `schema/` and `database/<engine>/` traverse AST nodes directly. So even if `./parser` were covered transitively by the engine jobs (it isn't, due to `-coverpkg=./...` not being enough on its own when the test set doesn't drive those code paths), the new `Format` methods would still register as 0% from cmd-rooted tests.

Filing a follow-up to add a `test-core` job to the workflow so parser-only patches get scored correctly going forward; that's an independent infra change and intentionally not bundled into this PR.

PR #1253, opened 2026-06-19
Fix pgquery multi-arg bool expressions

tsukuneA1:
## What changed

- Fold all `pgquery` `BoolExpr.Args` into sqldef's binary `AndExpr` / `OrExpr` AST nodes.
- Add a regression test for a `CHECK` constraint with three `OR` branches.

## Why

`pg_query_go` represents chained boolean expressions such as `a OR b OR c` as a single `BoolExpr` with multiple `Args`. The pgquery parser path previously converted only the first two args, which dropped later predicates from parsed expressions.

This could lose parts of `CHECK` constraints when sqldef falls back to the pgquery parser.

## Validation

```sh
GOCACHE=/tmp/sqldef-gocache go test ./database/postgres -run '^TestParse'
```


gfx:
@tsukuneA1 Thank you for contribution.


FYI I want to retire the pgquery parser once the generic parser can parse most of Pg DDLs.

PR #1260, opened 2026-06-30
psqldef: convert set operations and RangeSubselect in pgquery fallback

asfrgrtgd:
## What

- Add set operation conversion to `parseSelectStmt`: recursively parse `Larg`/`Rarg` into `*parser.Union` when `stmt.Op != SETOP_NONE` (UNION [ALL] / INTERSECT [ALL] / EXCEPT [ALL])
- Add `RangeSubselect` conversion to `parseSelectStmt`: wrap FROM clause subqueries in `*parser.Subquery`, preserving alias column names (`AS s(a, b)`) in `AliasedTableExpr.Columns`
- Fix double parentheses in `AliasedTableExpr.Format`: both `Format` and `Columns.Format` were emitting parentheses
- Fix `normalizeTableExpr` not copying `AliasedTableExpr.Columns` during reconstruction
- Fix no-FROM SELECT producing a single `From` entry with an empty `TableName`
- Return an explicit error for multiple FROM entries (`FROM a, b`) instead of silently dropping them

## Why (Motivate)

When the generic parser falls back to pgquery (e.g. `WITH (fillfactor = 70)`), VIEW definitions containing set operations or derived-table subqueries fail to parse.

```sql
CREATE VIEW v AS SELECT * FROM ((SELECT 1 AS value) UNION ALL (SELECT 2 AS value)) t;
```
parseSelectStmt rejected Op != SETOP_NONE and RangeSubselect, so SQL that the generic parser cannot handle failed in all modes.

Test

- database/postgres/tests.yml: add UNION ALL / UNION / INTERSECT / EXCEPT view tests with compare_with_generic_parser: true
- database/postgres/parser_test.go:
  - Fully-parenthesized UNION ALL in derived table (pgquery mode)
  - Generic parser failure with WITH (fillfactor = 70) followed by auto fallback with set operation view
  - Multiple FROM entries return an explicit error
  - Alias column names without double parentheses
  - Pin TestParseIndexAsync to auto mode (existing bug: failed under PSQLDEF_PARSER=pgquery)
- schema/generator_test.go: verify AS s(a, b) is preserved after normalization

```
PSQLDEF_PARSER=pgquery go test ./database/postgres
PSQLDEF_PARSER= go test ./database/postgres
go test ./schema
```


database/postgres/parser.go
```diff
@@ -368,20 +368,54 @@ func (p PostgresParser) parseViewStmt(stmt *pgquery.ViewStmt) (parser.Statement,
...
-	unhandled := stmt.IntoClause != nil ||
-		stmt.WindowClause != nil ||
-		stmt.SortClause != nil ||
-		stmt.ValuesLists != nil ||
+	if stmt.SortClause != nil ||
 		stmt.LimitOffset != nil ||
 		stmt.LimitCount != nil ||
-		stmt.LimitOption != 1 ||
-		stmt.LockingClause != nil ||
+		stmt.LimitOption != pgquery.LimitOption_LIMIT_OPTION_DEFAULT ||
```
asfrgrtgd, line 374: Replace magic number constants (`LimitOption != 1`, `Op != 1`) with named pgquery constants (`LIMIT_OPTION_DEFAULT`, `SETOP_NONE`)

database/postgres/parser.go
```diff
@@ -368,20 +368,54 @@ func (p PostgresParser) parseViewStmt(stmt *pgquery.ViewStmt) (parser.Statement,
...
+			}
+		default:
+			return nil, fmt.Errorf("unsupported set operation in parseSelectStmt: %d", stmt.Op)
+		}
+		return &parser.Union{Type: opType, Left: left, Right: right}, nil
+	}
+
+	if stmt.IntoClause != nil ||
+		stmt.WindowClause != nil ||
+		stmt.ValuesLists != nil {
```
asfrgrtgd, line 418: Move SELECT-only guard clauses (`IntoClause`, `WindowClause`, `ValuesLists`) after the set operation branch, since they do not apply to UNION/INTERSECT/EXCEPT

asfrgrtgd:
thanks for reviewing!

PR #1275, opened 2026-07-07
psqldef: convert JoinExpr in pgquery fallback

asfrgrtgd:
## What

- Extract FROM clause conversion from `parseSelectStmt` into `parseFromItem` for recursive FROM item handling, and add `JoinExpr` conversion via `parseJoinExpr`
  - `JOIN_INNER` / `JOIN_LEFT` / `JOIN_RIGHT` map to `join` / `left join` / `right join` (with `natural` variants when `is_natural` is set)
  - `JOIN_INNER` without quals/usingClause maps to `cross join` (pg_query has no dedicated CROSS type)
  - `USING (col, ...)` converts to `JoinCondition.Using`; `ON <expr>` delegates to `parseJoinQuals`
  - Right-side `JoinExpr` is wrapped in `ParenTableExpr` to preserve join associativity
  - Unsupported constructs (`JOIN_FULL`/`JOIN_SEMI`/`JOIN_ANTI`, aliased joins, `JOIN USING ... AS`) return errors, triggering generic-parser fallback in auto mode

- Refactor expression parsers (`parseBoolExpr`, `parseAExpr`, `parseFuncCall`, `parseNullTest`, `parseTypeCast`, `parseBooleanTest`, `parseCoalesceExpr`, `parseCaseExpr`) to accept a `parseOperand` callback, enabling reuse from both `parseExpr` and the new `parseJoinQuals`

- Add `parseJoinQuals` for JOIN ON conditions: unlike `parseExpr` which strips table qualifiers from column references, `parseJoinQuals` preserves them via `parseQualifiedColumnRef` (e.g., `ON u.id = p.user_id`) to avoid ambiguity

## Why (Motivate)

When the generic parser falls back to pgquery, VIEW definitions with JOIN in the FROM clause fail. `parseSelectStmt` only handles `RangeVar` and `RangeSubselect`, rejecting `JoinExpr`:

```sql
CREATE VIEW v AS SELECT * FROM (
  (SELECT a.id, a.name FROM a JOIN x USING (id))
  UNION ALL
  (SELECT b.id, b.name FROM b JOIN x USING (id))
) t;
```

In auto mode, pgquery conversion fails on `JoinExpr`, and the per-statement generic parser retry fails on the parenthesized set operation, aborting the migration with `syntax error ... near 'union'`.

Also improves coverage for `pg_get_viewdef()`-style definitions that use explicit join syntax.

## Test

- `database/postgres/tests.yml` (`compare_with_generic_parser: true`): covering JOIN USING (single/multiple columns), JOIN ON (simple/qualified/multiple conditions), INNER/LEFT/LEFT OUTER/RIGHT/NATURAL/NATURAL LEFT/NATURAL RIGHT/CROSS JOIN, nested joins (left/right), table aliases, ON with function calls/IS NULL/IS NOT NULL/type casts/boolean tests/COALESCE

- `database/postgres/parser_test.go`:
  - `TestJoinExprVariantsWithPgquery`: pgquery-mode cases including UNION ALL of join branches and right-nested parenthesized joins (no generic-parser comparison: pg_query doesn't preserve join parentheses while the generic parser uses `ParenTableExpr`)
  - `TestJoinExprErrorsWithPgquery`: FULL JOIN, aliased joins, `JOIN USING ... AS`, lateral subquery in join return conversion errors
  - `TestJoinExprErrorsWithPgqueryNodes`: error paths via direct node construction (SEMI join, invalid FROM items, bad USING/ON nodes, qualified star column, excess column ref fields)

```
PSQLDEF_PARSER=pgquery go test ./database/postgres
PSQLDEF_PARSER= go test ./database/postgres
go test ./schema
go test ./cmd/psqldef
```

### DIFF
--- a/database/postgres/parser.go
+++ b/database/postgres/parser.go
@@ -547,9 +547,85 @@ func (p PostgresParser) parseFromItem(node *pgquery.Node) (parser.TableExpr, err
 			TableHints: []string{},
 			As:         parser.NewIdent(aliasName, false),
 		}, nil
+	case *pgquery.Node_JoinExpr:
+		return p.parseJoinExpr(n.JoinExpr)
 	default:
 		return nil, fmt.Errorf("unknown node in parseFromItem: %#v", n)
 	}
+}
+
+func (p PostgresParser) parseJoinExpr(join *pgquery.JoinExpr) (parser.TableExpr, error) {
+	if join.JoinUsingAlias != nil {
+		return nil, fmt.Errorf("unhandled join using alias in parseJoinExpr: %#v", join.JoinUsingAlias)
+	}
+	if join.Alias != nil {
+		return nil, fmt.Errorf("unhandled join alias in parseJoinExpr: %#v", join.Alias)
+	}
+
+	var joinType string
+	switch join.Jointype {
+	case pgquery.JoinType_JOIN_INNER:
+		if join.IsNatural {
+			joinType = parser.NaturalJoinStr
+		} else if join.UsingClause == nil && join.Quals == nil {
+			// pgquery has no dedicated CROSS join type: CROSS JOIN is
+			// represented as JOIN_INNER without a join condition
+			joinType = parser.CrossJoinStr
+		} else {
+			joinType = parser.JoinStr
+		}
+	case pgquery.JoinType_JOIN_LEFT:
+		if join.IsNatural {
+			joinType = parser.NaturalLeftJoinStr
+		} else {
+			joinType = parser.LeftJoinStr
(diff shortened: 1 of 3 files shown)


gfx:
Can you fix the generic parser?

pgquery is about to be retired.

asfrgrtgd:
@gfx OK! I'm gonna fix generic parser and not gonna change pgquery mainly.

asfrgrtgd:
closed this pr because i'm creating  new pr to make it work with the generic parser instead of pgquery fallback

PR #1285, opened 2026-07-21
feat(psqldef): extend managed_roles privilege coverage (WITH GRANT OPTION, sequences, columns)

dip-daiki-ogikubo:
## Summary

`managed_roles` currently manages table-level GRANTs only. Everything else
requires a second, hand-rolled privilege pipeline outside psqldef:

- `WITH GRANT OPTION` is explicitly rejected by both parsers
  ("WITH GRANT OPTION is not supported yet")
- `GRANT ... ON SEQUENCE` never reaches the grammar because the tokenizer
  mapped `sequence` (and `usage`) to the catch-all UNUSED token
- column-level privileges (`GRANT SELECT (col) ON TABLE ...`) parse in the
  pgquery fallback but the column list is silently dropped

This PR closes those three gaps so real-world privilege sets (e.g. app roles
with sequence USAGE for SERIAL columns, audit roles with column-restricted
SELECT) can be managed declaratively.

It is also the parsing/diff/export engine that the `manage.privilege` section
of the object-management RFC (#1072) needs in order to manage GRANTs on
sequences and columns. Per #1278, the gaps are closed in the generic parser
(`parser/parser.y`) rather than by relying on the pgquery fallback — every new
form is verified with `PSQLDEF_PARSER=generic` (the pgquery converter is also
updated to match, so a whole-file fallback doesn't reject these statements).

## Changes

- **WITH GRANT OPTION**: thread the flag through parse → diff → export.
  Export reads `is_grantable` from `information_schema.table_privileges`.
  The diff also handles in-place changes: re-GRANT to add the option
  (idempotent in PostgreSQL) and `REVOKE GRANT OPTION FOR ...` to remove it
  (gated by `enable_drop` like other REVOKEs).
- **Sequence privileges**: promote `SEQUENCE` from UNUSED to a real keyword
  and add `GRANT/REVOKE ... ON SEQUENCE` grammar (incl. WITH GRANT OPTION).
  `Grant.ObjectType` distinguishes tables from sequences end to end.
  Export emits sequence grants for managed roles from `pg_class.relacl`
  (`relkind='S'`), excluding extension-owned sequences. `ALL` expands to the
  sequence privilege list (SELECT, UPDATE, USAGE).
- **Column privileges**: `privilege` accepts a parenthesized column list.
  Privileges are canonicalized (keyword uppercased, columns sorted, quoted
  only when necessary) so declaration order and redundant quoting do not
  cause spurious diffs. Export reads explicit column grants from
  `pg_attribute.attacl` (unlike `information_schema.column_privileges`, it
  does not repeat privileges implied by table-level grants).
- **pgquery fallback**: accept `OBJECT_SEQUENCE` and convert
  `AccessPriv.Cols` instead of dropping them.

## Tokenizer compatibility

`sequence` was previously usable as an identifier via the UNUSED token, so
`table_id` now accepts the `SEQUENCE` keyword to keep tables named
"sequence" parsing (PostgreSQL treats SEQUENCE as an unreserved keyword),
following the TRIGGER/INDEX precedent in #1282. `usage` stays mapped to
UNUSED; the `privilege` rule accepts UNUSED keywords, which matches the
grammar's existing permissiveness for privilege names. goyacc conflicts stay
at 29 shift/reduce (+1 reduce/reduce).

## Verification

- New round-trip tests: sequence grant / WGO / idempotency / drift revoke /
  orphan revoke, column grant / idempotency / mixed with table grants /
  drift revoke, WGO add / upgrade / downgrade / idempotency
- pgquery-mode unit tests for the GRANT conversion (no DB required)
- Full `make test-psqldef` on PostgreSQL 13, 18, and pgvector
  (`PSQLDEF_PARSER=generic` and the default parser)
- `make test-mysqldef` on MySQL 8.0/8.4/9, MariaDB 11/12, TiDB v8.5.4 and
  `make test-sqlite3def` pass (the grammar/tokenizer are shared across
  dialects)
- As a real-world acceptance check, all 2,802 GRANT statements from a
  production privilege catalog (221 of them sequence grants, plus column
  grants) parse and round-trip
- `make parser` reproduces the committed parser.go; `make lint` clean


dip-daiki-ogikubo:
Rebased onto the latest master to resolve the conflict with #1291 (the merge-queue run hit it — #1291 landed right before this PR was queued, touching the same parser files). The regenerated `parser/parser.go` now incorporates the COMMENT-as-column-name change, and the full local test matrix (psqldef PG13/18 generic, mysqldef 8.0, sqlite3def, core, vet) is green. Should be ready for the queue again — sorry for the extra round trip!

PR #1286, opened 2026-07-22
psqldef: support operator classes on expression index columns

moznion:
The generic parser's index_column rule enumerated the index element grammar by hand, but only covered operator classes on plain column references. Expressions had no operator class variant, so the standard pg_get_indexdef output

```
CREATE INDEX ... USING gin (((notes ->> 'text'::text)) gin_trgm_ops)
CREATE INDEX ... USING btree (lower(name) text_pattern_ops)
```

failed with a syntax error at the operator class name, which routed the whole file to the pgquery fallback. Any database with a pg_trgm expression index hits this on --export.

Add the two missing productions. Adding the function_call_generic one alone introduces a reduce/reduce conflict on TEXT_PATTERN_OPS: that token is the only operator class name lexed as a keyword instead of an identifier, and it lands in a state where function_call_generic already conflicts with the empty over_expression rule. Operator class names are just identifiers, so drop the dedicated token and let operator_class: sql_id handle it. Conflict counts are unchanged at 29 shift/reduce and 542 reduce/reduce.

### Why dropping TEXT_PATTERN_OPS is safe:

- The token had exactly three references: the %right declaration, the single operator_class alternative, and the keyword map entry. No rule used it via %prec, and no Go code outside the generated parser referred to it. Removing a whole precedence level does not reorder the remaining ones.

- operator_class: sql_id has handled every other operator class name since bd9cb61 added that alternative. The dedicated token only survived for historical reasons: when 12d8687 introduced it, operator_class had no generic alternative at all. varchar_pattern_ops and gin_trgm_ops have always gone through sql_id.

- text_pattern_ops is not a PostgreSQL keyword. It is a row in pg_opclass, i.e. an ordinary identifier. Reserving it was a defect of its own: the parser rejected both CREATE TABLE text_pattern_ops (...) and CREATE TABLE t (text_pattern_ops text), which PostgreSQL accepts. IndexOnIdentifierNamedTextPatternOps covers this.

- A differential test parsed all 5793 statements of the cmd/*/tests*.yml corpus, across all four dialects, with the parser built before and after this change. Every difference is ERR -> OK and belongs to a test case added here; nothing regressed from OK to ERR.

One behavioral difference remains, and it is a fix. The keyword path lowercased the name, so `name TEXT_PATTERN_OPS` parsed to the operator class `text_pattern_ops`; the identifier path preserves it as written, which is what varchar_pattern_ops and the rest already did. The name is emitted unquoted, so PostgreSQL case-folds it: applying `CREATE INDEX idx_up ON t (name TEXT_PATTERN_OPS)` yields `USING btree (name text_pattern_ops)` in the database and the next run reports no changes.


gfx:
Conflict is happining 🙏

moznion:
@gfx Thank you for your notice. I just resolved the conflict.

PR #1289, opened 2026-07-22
psqldef: keep NULLS FIRST/LAST on index columns

moznion:
This depends on https://github.com/sqldef/sqldef/pull/1288; the actual changes are https://github.com/sqldef/sqldef/commit/e1d525dd6b75624bc83037e11af43132f1f3ba98

---

schema.IndexColumn had no field for it, so the parser's NullsOrdering was dropped on the way in. Regenerating a statement in quote-aware mode then silently created an index that sorts NULLs the other way, and the comparison could not see it either, so no further DDL was ever produced.

Which end NULLs sort at is only stated in the DDL when it differs from the default of the direction, so compare the resolved ordering rather than the clause: an omitted one has to equal an explicit NULLS LAST on an ascending column.


moznion:
#1372 has already addressed the same issue.

PR #1292, opened 2026-07-26
psqldef: Fix unnamed composite unique names

takaram:
## Summary

- Normalize unnamed PostgreSQL composite `UNIQUE` constraints using every target column in declaration order
- Reuse the existing PostgreSQL identifier quoting and 63-byte truncation behavior
- Update the existing composite constraint DDL expectations and add an idempotency regression test

## Background

PostgreSQL derives the default name of an unnamed composite unique constraint from all constrained columns, such as `orders_priority_category_key`. sqldef only applied the PostgreSQL naming convention to single-column unique constraints and fell back to the first column name for composite constraints.

As a result, the desired schema and the schema exported after application used different constraint names, which could produce unnecessary `ADD CONSTRAINT` and `DROP CONSTRAINT` statements on subsequent runs.

This fixes the behavior described in Discussion #1290.

## Impact

Unnamed PostgreSQL composite unique constraints now remain idempotent after they are applied. Explicitly named constraints and the MySQL, SQLite, and SQL Server naming paths are unchanged.

## Validation

- Targeted PostgreSQL integration tests with `PSQLDEF_PARSER=generic`
- `make format`
- `make build`
- `make test-all-flavors`
- `make lint`

### DIFF
--- a/cmd/psqldef/tests_constraints.yml
+++ b/cmd/psqldef/tests_constraints.yml
@@ -920,11 +920,24 @@ ConstraintCheckInMultipleColumnsWithUnique:
   up: |
     ALTER TABLE "public"."orders" ADD CONSTRAINT "orders_priority_check" CHECK (priority = ANY (ARRAY[1, 2, 3]));
     ALTER TABLE "public"."orders" ADD CONSTRAINT "orders_category_check" CHECK (category = ANY (ARRAY[10, 20, 30, 40]));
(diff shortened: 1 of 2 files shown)


takaram:
This PR was removed from the merge queue due to what looks like a transient Docker communication error during CI.

@gfx 
Could you please retry it?

PR #1302, opened 2026-07-31
psqldef: update changed functions with CREATE OR REPLACE instead of DROP + CREATE

dip-daiki-ogikubo:
## Summary

When a function's body (or another non-signature property) changed, `generateDDLsForCreateFunction` emitted `DROP FUNCTION` + `CREATE` even though its own comment said it should "use CREATE OR REPLACE". The destructive drop-and-recreate loses dependent objects (triggers, views, column defaults), GRANTs and COMMENTs; and under `enable_drop: false` the `DROP` is skipped while the plain `CREATE` then fails with `already exists`, so the change never converges.

While fixing this I found and fixed two underlying bugs, so this PR has two commits.

## Commit 1 — update changed functions with CREATE OR REPLACE

For PostgreSQL, a changed function is now updated in place with `CREATE OR REPLACE` when its **signature** is unchanged, and only dropped-and-recreated when the signature changed (which PostgreSQL requires: changing the return type is an error, and changing argument types creates an overload instead of replacing).

- **Root cause of the redundant drop**: the generic grammar's `or_replace_opt` action returned `""` for `OR REPLACE`, so `Function.OrReplace` was never set and the `OR REPLACE` branch was dead code. The grammar now yields a non-empty value, making the flag reliable (no structural change — conflict counts are unchanged).
- **Signature identity** compares argument modes (`IN`/`OUT`/`INOUT`/`VARIADIC`), names (folded like identifiers — unquoted lower-cased, quoted preserved; adding a name to an unnamed parameter is allowed, renaming is not), and types normalized across common aliases (`int`/`integer`, `varchar`/`character varying`, …) since the current side comes from `pg_get_functiondef`.
- `RETURNS TABLE(...)` loses its column list in parsing, so it is always treated as non-replaceable (drop + create).
- **Overloads**: the changed function is paired with the current definition sharing its signature; when the pairing is ambiguous the previous behavior is kept. `DROP FUNCTION` now carries the identity argument types so it stays unambiguous under overloads.
- OR REPLACE is spliced in via a token check on the leading `CREATE` keyword (no regex SQL parsing); the decision that it's missing comes from the parsed `OrReplace` flag.
- Other dialects are unchanged.

## Commit 2 — keep overloaded functions distinct when ordering DDLs

`SortTablesByDependencies` keyed every object by its normalized name, so PostgreSQL functions overloaded on argument types collapsed to one node in `topologicalSort`: one representative was replayed for every slot, silently dropping the other overloads and duplicating one. This corrupted output whenever two same-named functions appeared together (reproducible on the current tree, independent of commit 1). Items sharing a key are now grouped, one representative per key is sorted, then each is expanded back to all its items in original order; single-item keys (the common case) are unaffected.

## Verification

- New unit tests (`areSameFunctionSignature`, `insertOrReplaceIntoCreateFunction`, `dropFunctionDDL`) and round-trip tests (`tests_functions.yml`): body change → CREATE OR REPLACE; return-type / argument-type / OR-REPLACE-with-return-type change → drop + recreate; `RETURNS TABLE` change → drop + recreate; overloaded functions both replaced in place.
- Verified on a live PostgreSQL that a body change preserves a dependent trigger, the function COMMENT, and a GRANT, and applies the new body.
- `test-core`, psqldef on PostgreSQL 13 / 18 / pgvector, mysqldef on MySQL 8.0 / 9 / MariaDB 12 / TiDB, sqlite3def, `go vet`, and `make parser` (deterministic, conflict counts unchanged) all pass locally.

### DIFF
--- a/cmd/psqldef/tests_functions.yml
+++ b/cmd/psqldef/tests_functions.yml
@@ -397,3 +397,79 @@ CreateFunctionWithDefaultArgs:
     $$ LANGUAGE plpgsql VOLATILE;
   up: ""
   down: ""
+
+# Changing only the body of an existing function uses CREATE OR REPLACE, so it
+# converges without dropping (and works even with enable_drop: false).
+FunctionBodyChangeUsesCreateOrReplace:
+  legacy_ignore_quotes: false
+  enable_drop: false
+  current: |
+    CREATE FUNCTION add_one(x integer) RETURNS integer AS $$ BEGIN RETURN x + 1; END; $$ LANGUAGE plpgsql;
+  desired: |
+    CREATE FUNCTION add_one(x integer) RETURNS integer AS $$ BEGIN RETURN x + 2; END; $$ LANGUAGE plpgsql;
+  up: |
+    CREATE OR REPLACE FUNCTION add_one(x integer) RETURNS integer AS $$ BEGIN RETURN x + 2; END; $$ LANGUAGE plpgsql;
+  down: |
+    CREATE OR REPLACE FUNCTION add_one(x integer) RETURNS integer AS $$ BEGIN RETURN x + 1; END; $$ LANGUAGE plpgsql;
+
+# Changing the return type cannot go through CREATE OR REPLACE (PostgreSQL
+# rejects it), so the function is dropped and recreated.
+FunctionReturnTypeChangeDropsAndRecreates:
+  legacy_ignore_quotes: false
+  current: |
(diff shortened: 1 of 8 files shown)


dip-daiki-ogikubo:
Heads-up: the merge-queue run that dropped this PR (and the others in the same batch) failed only on `test (mysqldef, mysql, 8.4)` with repeated `[mysql] packets.go:58 unexpected EOF` — a MySQL container connection flake, not a test assertion. This PR's own branch checks are green and it's `MERGEABLE`/`CLEAN` (PostgreSQL-only change: parser + generator). Happy to re-queue whenever convenient; let me know if you'd like me to rebase or split the two commits.

PR #1304, opened 2026-07-31
psqldef: parse column references inside ARRAY[...] and keep their quotes

178inaba:
## Why

PostgreSQL lets `ARRAY[...]` elements be column references, but the generic parser rejects them:

```sql
CREATE TABLE t (
    "Status" text,
    "Fallback" text,
    CONSTRAINT t_status_check CHECK ("Status" = ANY (ARRAY["Fallback", 'pending']))
);
```

All three parser modes fail on this today:

| `PSQLDEF_PARSER` | before |
|---|---|
| unset (auto) | generic fails → pgquery fallback → `unknown expr in parseArrayElement` |
| `pgquery` | same error |
| `generic` | `syntax error at line 1, column 102 near 'Fallback'` |

`AGENTS.md` treats a pgquery fallback as a bug in itself. Auto mode routes the whole file to pgquery when the generic parser fails on any single statement, so one such table drags unrelated ones along with it. `v4-migration.md` Breaking Change 3 removes the fallback in v4, which makes a generic-parser gap a v4 blocker as well.

**Fixing the parse alone would generate wrong DDL.** `formatExprQuoteAware` has no `ArrayConstructor` branch, so it falls back to `parser.String`, which ignores `Ident.Quoted` and drops the quoting. PostgreSQL folds an unquoted identifier to lower case, so `ARRAY["Fallback", ...]` would come out as `ARRAY[Fallback, ...]` and reference a different column — an error at apply time in most cases, but a silently different constraint when both `Fallback` and `fallback` exist. That is why both changes ship in the same PR.

Found while reviewing #1295, which added `ANY`/`ALL` output to `formatExprQuoteAware` and thereby made `ARRAY[...]` reachable on the quote-aware path. It was recorded there as a FIXME in `schema/normalize_any_array_test.go`; this PR enables that case.

## What

1. `parser/parser.y` — add `column_name` to `array_element`. `array_constructor` has a single form (`ARRAY '[' array_element_list ']'`), so this covers every context an `ARRAY[...]` can appear in (CHECK constraints, index expressions, DEFAULT).
2. `schema/generator.go` — add the `ArrayConstructor` case to `formatExprQuoteAware`, formatting elements recursively.
3. Tests — enable the FIXME case, add parser-level cases, add an end-to-end YAML case.

`column_name` rather than the narrower `sql_id`: `non_reserved_keyword` holds type names and general words (`TIME`, `TIMESTAMP`, `VALUE`), not special-value keywords like `CURRENT_USER`, so there is no risk of reading a special value as a column reference. PostgreSQL also treats `value` as non-reserved and usable as a column name.

**Conflicts match the master baseline at 28 shift/reduce, 545 reduce/reduce — this PR adds none.** `AGENTS.md` says no reduce/reduce conflicts are allowed while the tree already carries 545; this PR reads that as "no new conflicts" and does not attempt to reduce the existing count.

## Known limitation (deliberately not fixed here)

**Under `legacy_ignore_quotes: true` (the current default), a column reference inside `ARRAY[...]` is still not emitted correctly.** `normalizeCheckExprString` takes the `parser.String` branch in legacy mode, so the `ArrayConstructor` case does not apply and the quotes are still dropped.

This is not specific to `ARRAY`; it is the existing behaviour of legacy mode for CHECK expressions generally (`CHECK ("Status" <> 'x')` already emits `Status <> 'x'`, and `lower("Status")` emits `lower(Status)`).

So this PR changes legacy-mode behaviour from "a clear parse error" to "DDL that PostgreSQL rejects". Reasons for leaving it:

1. There is no correct behaviour for legacy mode. Always quoting breaks unquoted mixed case (`CREATE TABLE Users` means `users` in PostgreSQL); never quoting breaks quoted mixed case. That is exactly why v4 flips the default.
2. Changing legacy output would rewrite the emitted text of every existing CHECK constraint, unrelated to `ARRAY` — a very different risk profile from a grammar addition with zero new conflicts.
3. `v4-migration.md` states the option "will be removed in a future major release".

## Verification

<details>
<summary>Round-trip against PostgreSQL 18</summary>

```console
$ psql -d roundtrip_check
=# CREATE TABLE t ("Status" text, "Fallback" text,
       CONSTRAINT t_status_check CHECK ("Status" = ANY (ARRAY["Fallback", 'pending'])));
CREATE TABLE

$ psqldef --config-inline "legacy_ignore_quotes: false" roundtrip_check --export
CREATE TABLE public.t (
    "Status" text,
    "Fallback" text,
    CONSTRAINT t_status_check CHECK ("Status" = ANY (ARRAY["Fallback", 'pending'::text]))
);

$ psqldef --config-inline "legacy_ignore_quotes: false" roundtrip_check --dry-run --file schema.sql
-- Nothing is modified --
```

Auto mode (`PSQLDEF_PARSER` unset) emits no `Generic parser failed ... using pgquery fallback` warning for this schema.

</details>

<details>
<summary>Regression guards for the other array_element alternatives</summary>

Adding an alternative to `array_element` could change how the existing `typed_literal` / `current_timestamp` branches resolve. These existing cases were checked explicitly and still pass:

- `TestApply/NumericTypedLiteralInArrayDefault` — `ARRAY[numeric '1']`
- `TestApply/UuidJsonTypedLiteralInCheck` — `uuid` / `json` typed literals
- `TestApply/CreateTableWithDefault` — `ARRAY[current_date::text]::text[]`
- `TestApply/CheckConstraintWithTypeCastInArray` — `ARRAY[...]` with element casts
- `TestApply/ConstraintCheckArrayColumnReferenceMixedTypesQuoteAware` — column reference travelling through the array-level cast PostgreSQL emits for mixed column types (`(ARRAY[...])::text[]`)

Keyword-named columns such as `ARRAY[time, 'x']` remain a syntax error, since the existing `typed_literal: TIME STRING` branch wins. That matches the baseline and is not a regression.

</details>

## Out of scope

- The same gap in the pgquery conversion path (`parseArrayElement` in `database/postgres/parser.go`). `PSQLDEF_PARSER=pgquery` is a development-only setting that v4 removes, and auto mode re-parses only the statements pgquery fails to convert, so fixing the generic parser is what matters.
- Retiring `formatExprQuoteAware` itself — it is a hand-written parallel implementation of `parser.String`, which is why cases keep getting missed. Worth a separate change.
- Arbitrary expressions inside `ARRAY[...]` (`ARRAY[lower(col)]`, `ARRAY[col || 'x']`). Replacing `array_element` with a general expression collides head-on with the `typed_literal` / `tuple_expression` / `TYPECAST` branches and needs its own investigation. A bare cast on a column reference (`ARRAY["Fallback"::text]`) is still a syntax error, while the parenthesised form PostgreSQL exports (`ARRAY[("Fallback")::text]`) parses — worth knowing as the starting point for the next slice. This PR is a partial fix for the generic-parser gap.
- Legacy-mode quote handling inside expressions, per the limitation above.


gfx:
Please resolve the conflict 🙏

178inaba:
@gfx 
Rebased onto the latest master.
The conflict is resolved and CI is green 🫡

moznion:
@178inaba Sorry for bothering you again. The changes look good, so I think this is ready to merge. Could you please resolve the merge conflict by regenerating the parser?

178inaba:
@moznion 
I’ve rebased it! Thanks!

moznion:
I appreciate your contribution.

PR #1308, opened 2026-08-01
parser: drop duplicate NEWID/NEWSEQUENTIALID default value rules

178inaba:
## Why

`default_value_expression` already includes `function_call_nonkeyword`, which defines `NEWID '(' ')'` and `NEWSEQUENTIALID '(' ')'`. It also redefines both of them directly, with identical actions. The duplication causes 60 reduce/reduce conflicts, 30 each in states 1669 and 1670, and `AGENTS.md` says no reduce/reduce conflicts are allowed.

## What

Drop the two duplicate alternatives from `default_value_expression`, leaving the `function_call_nonkeyword` ones as the only path. Both build `&FuncExpr{Name: NewIdent($1, false)}`, so the accepted SQL and the resulting AST are unchanged. `make parser` reports reduce/reduce dropping from 545 to 485, with shift/reduce unchanged at 28.

Two regression cases for `DEFAULT NEWID()` and `DEFAULT NEWSEQUENTIALID()` are added to `TestDefaultFunctionExpressions`.


moznion (approved):
@178inaba Thank you for your work. This seems fine so this should be merge ready but the generated parser has a conflict. Could you please resolve that?

178inaba:
@moznion 
Thank you! I rebased the branch and resolved the conflicts.

CodeQL is failing, but it seems to be caused by a temporary issue on GitHub's side.
I pushed an empty commit and confirmed that it passes:
https://github.com/sqldef/sqldef/actions/runs/34748855904

Since this has been approved, I'll go ahead and merge it.
If anything goes wrong, please let me know and I'll revert it.

PR #1321, opened 2026-08-03
mysqldef: Add --disable-ddl-transaction flag

zaneli:
I've been using an old version (v0.11) of mysqldef, and after upgrading to the latest version I noticed that BEGIN/COMMIT are now added around DDL statements.
I'd like to be able to turn this off via a setting if possible (currently I'm working around it by stripping those lines with `grep -v -E`).

It looks like #1177 added this flag to psqldef only, so I'd like to add it to mysqldef as well, if that's acceptable.


moznion (approved):
Looks nice, thank you for your contribution.

PR #1329, opened 2026-08-12
psqldef: accept a parenthesized comparison as a comparison operand

dim0627:
## Summary

The generic parser rejects a `CHECK` whose left operand is a parenthesized comparison:

```sql
CREATE TABLE items (
  format text NOT NULL,
  interval_ms integer,
  CONSTRAINT items_interval_check CHECK ((format = 'animated'::text) = (interval_ms IS NOT NULL))
);
```

```
syntax error at line 4, column 74
```

This is not an exotic shape: it is a natural way to express an "if and only if" invariant between two columns, and PostgreSQL stores and returns it verbatim through `pg_get_constraintdef()`, so it round-trips out of `--export` and straight back into the desired schema.

The failure was oddly selective, which is what made it worth digging into:

| CHECK expression | before |
|---|---|
| `((a = 'x'::text) = (b IS NOT NULL))` | syntax error |
| `((b IS NOT NULL) = (a = 'x'::text))` | OK |
| `((a = 'x'::text) = flag)` | syntax error |
| `(flag = (a = 'x'::text))` | OK |
| `((b > 1) = (b < 5))` | syntax error |

## Cause

Two rules match a parenthesized condition:

```
condition:  '(' condition ')'
row_tuple:  '(' expression_list ')'
```

`y.output` shows both items live in the same state, and yacc resolves the reduce/reduce conflict in favour of the rule that appears earlier in the grammar — `condition`. So `(a = 'x')` reduces to a `condition`, and a `condition` cannot be an operand of `compare`, whose rule is `value_expression compare value_expression`.

A parenthesized `IS NOT NULL` is an `expression` rather than a `condition`, so it never matches the first rule, takes the `row_tuple` path, and reduces through `tuple_expression` to `value_expression` — which is why it parses in the same position. That asymmetry is the whole bug.

## Fix

Both rules build the same node:

```go
// condition: '(' condition ')'
$$ = &ParenExpr{Expr: $2}

// tuple_expression: row_tuple
if len($1) == 1 {
    $$ = &ParenExpr{$1[0]}
}
```

so dropping `condition: '(' condition ')'` leaves the AST unchanged and lets every parenthesized condition take the `row_tuple` path, which reduces to `value_expression` and can therefore appear on either side of a comparison.

Conflict counts are unchanged: **28 shift/reduce, 545 reduce/reduce**, before and after. The `%prec LOWER_THAN_RPAREN` comment that named the removed rule is updated to name the rule the conflict is actually against now.

## Tests

- `parser/parser_test.go`: `TestParenthesizedComparisonAsComparisonOperand` covers the shapes in the table above, plus a case asserting a plain parenthesized condition still yields a `ParenExpr` so the AST did not change.
- `cmd/psqldef/tests_constraints.yml`: `CheckConstraintComparingParenthesizedComparison` exercises add/drop of such a constraint against a live PostgreSQL. Per `AGENTS.md` this runs under `PSQLDEF_PARSER=generic` in CI, which is the documented way to verify that a fix removed a fallback rather than relying on one.

Both fail on `master` and pass here.

## Verification

Run against real servers, since the grammar is shared by all four commands:

| suite | result |
|---|---|
| `make test-core` | 340 tests |
| `make test-sqlite3def` | 82 tests |
| `make test-mysqldef` (MySQL 8.0) | 332 tests, 27 skipped |
| `make test-psqldef` (PostgreSQL 18, `PSQLDEF_PARSER=generic`) | 968 tests, 17 skipped |
| `make test-mssqldef` (SQL Server 2022) | 102 subtests |

`make build`, `make lint` and `make format` are clean, and `make parser` regenerates `parser/parser.go` reproducibly.

MariaDB, TiDB and pgvector flavors were not run locally.

## Note

Found while trying to move a schema from psqldef v0.17.29 to v3 (discussion #1328). That report describes the failure as needing a `GENERATED ALWAYS AS IDENTITY` column as well, because the identity column is what stopped the pgquery fallback from rescuing the file. The grammar gap itself is independent of identity columns, and this fixes it on the generic parser side, per the `AGENTS.md` policy that a fallback to pgquery is itself a bug.


moznion:
@dim0627 Thank you for your contribution. It seems good. Could you please resolve the conflicts?

dim0627:
@moznion Thanks for the review! Resolved the conflicts and pushed. Could you take another look?

moznion:
@dim0627 Thank you for your response, but it looks like parser/parser.go still has a merge conflict. Could you please take a look?

moznion:
I appreciate your work!

PR #1330, opened 2026-08-15
psqldef: accept unquoted key as a column reference in the generic parser

178inaba:
## Why

`key` is an unreserved keyword in PostgreSQL, so it works as an unquoted
column name. #1202 introduced the `PG_KEY` token so column *declarations*
(`key text NOT NULL`) and index column lists parse, but the column
*reference* paths were left without a slot: `column_name` for bare
references in expressions, and `reserved_sql_id` for the right-hand side of
a qualified name.

As a result these fail to parse in the generic parser:

```sql
COMMENT ON COLUMN t.key IS 'a key';
CREATE VIEW v AS SELECT key FROM t;
CREATE VIEW v AS SELECT id FROM t WHERE key = 'a';
CREATE INDEX i ON t (key) WHERE key IS NOT NULL;
CREATE TABLE t (key int, CHECK (key > 0));
```

With `PSQLDEF_PARSER=generic` this is a hard error. In the default auto mode
the *whole file* falls back to pgquery, which AGENTS.md calls a bug in
itself. It also hits the current side: a view that selects `key` fails again
when its definition is read back from the database and re-parsed.

## What

Add `PG_KEY` alternatives to `column_name` and `reserved_sql_id`, exactly as
#1291 did for `PG_COMMENT`. `PG_KEY` now appears in the same four sites as
`PG_COMMENT` (`column_definition`, `col_name_keyword`, `column_name`,
`reserved_sql_id`).

`new_qualifier_column_name` is `NEW '.' reserved_sql_id` only, so `NEW.key`
in a trigger condition is covered by the `reserved_sql_id` slot.

Conflicts are unchanged from the base commit: **28 shift/reduce,
487 reduce/reduce**.

### Why not route these through `col_name_keyword`

`col_name_keyword` already groups `DATE | KEY | PG_KEY | PG_COMMENT | TIME |
TIMESTAMP | VALUE` and is referenced from the index and alias rules, so it
looks like the general fix. It is not a drop-in here: `column_name` and
`reserved_sql_id` currently do not accept bare `DATE`, `TIME`, `TIMESTAMP`
or `KEY` as identifiers, and switching them to `col_name_keyword` would
widen what they accept well beyond `key`. That is a separate design call
about the accepted identifier set, not part of this fix.

### Column reference positions surveyed

Verified against the generic parser in Postgres mode. Now parsing:

| Position | Example |
|---|---|
| `COMMENT ON COLUMN`, qualified | `COMMENT ON COLUMN t.key IS '...'` |
| `COMMENT ON COLUMN`, schema-qualified | `COMMENT ON COLUMN public.t.key IS '...'` |
| Select list, bare and qualified | `SELECT key FROM t` / `SELECT t.key FROM t` |
| `WHERE` | `SELECT id FROM t WHERE key = 'a'` |
| Partial index predicate | `CREATE INDEX i ON t (key) WHERE key IS NOT NULL` |
| `CHECK` | `CREATE TABLE t (key int, CHECK (key > 0))` |
| `REFERENCES` column list | `CREATE TABLE t2 (id int REFERENCES t (key))` |
| Trigger `UPDATE OF` | `CREATE TRIGGER ... AFTER UPDATE OF key ON t ...` |
| Trigger `WHEN`, `NEW`-qualified | `... FOR EACH ROW WHEN (NEW.key IS NOT NULL) ...` |
| Column-level `GRANT` | `GRANT SELECT (key) ON t TO r` |
| `GROUP BY` / `ORDER BY` | `SELECT key FROM t GROUP BY key` |

`INSERT INTO t (key)`, `ON CONFLICT (key)` and `ALTER TABLE ... RENAME
COLUMN key` remain unparseable, but that is unrelated to `key` — the generic
parser has no grammar for those statements at all (they fail identically
with a non-keyword column name). For the record, `ins_column` does not go
through `reserved_sql_id` and so does not accept `PG_KEY`; it is only
reachable from a trigger body, which PostgreSQL triggers do not have.

## Tests

- `parser/parser_test.go`: `TestKeyKeywordAsColumnReference` covers every
  position above, the declaration positions that already worked, and both
  directions of the MySQL split — inline `KEY idx (col)` / `UNIQUE KEY`
  still parse, and bare `key` as a column reference is still rejected.
- `cmd/psqldef/tests_comments.yml`: `CommentOnColumnNamedKey`.
- `cmd/psqldef/tests_views.yml`: `ViewReferencingColumnNamedKey`, which
  exercises the read-back path where `pg_get_viewdef()` output is re-parsed
  as the current state.

Verified with `make test-all-flavors` (MySQL, MariaDB, TiDB, PostgreSQL,
pgvector, SQL Server, SQLite3), plus `make build`, `make format` and
`make lint`.


moznion (approved):
Thank you!

PR #1333, opened 2026-08-20
psqldef: support CREATE CONSTRAINT TRIGGER (deferrable triggers)

ceftx:
PostgreSQL's deferrable constraint triggers (CREATE CONSTRAINT TRIGGER ... DEFERRABLE INITIALLY DEFERRED) weren't recognized anywhere in the generic parser's CREATE TRIGGER grammar, so both --dry-run (reading one from schema.sql) and --export (reconstructing one from pg_get_triggerdef() output during introspection) hard-failed the moment such a trigger existed anywhere in the target schema — not just for that one object, breaking the whole run.

Adds an optional CONSTRAINT keyword and an optional deferrable clause (reusing the existing deferrable_option production already used for FK/ UNIQUE constraints) to the two FOR EACH ROW EXECUTE FUNCTION/PROCEDURE trigger productions, since constraint triggers are always AFTER + FOR EACH ROW and never take a WHEN clause. The FROM referenced_table_name clause isn't supported yet.

Threads Constraint/ConstraintOptions through the Trigger AST node into schema.Trigger, generates the CONSTRAINT keyword and an explicit DEFERRABLE/ NOT DEFERRABLE clause in generateDDLsForCreateTrigger (Postgres reconstructs CREATE TRIGGER DDL from structured fields rather than echoing the original statement text), and extends areSameTriggerDefinition to compare constraint options so a deferrable-option-only change is detected as a real diff.


moznion (commented):
@ceftx Unfortunately, this pull request has a merge conflict. Could you please resolve it?

ceftx:
@moznion good, I'm on it

ceftx:
@moznion ready brou

PR #1334, opened 2026-08-20
psqldef: export GRANT privileges for views and materialized views

dip-daiki-ogikubo:
## Problem

Table and column privileges for ordinary tables are attached to the table DDL via `getPrivilegeDefsForTables`, whose object list comes from `tableNames()` (`relkind in ('r','p')`). Views and materialized views are dumped through a separate path (`views()` / `materializedViews()`) that carries no privileges, so **their GRANTs were never exported**.

Under `manage.privilege` (or the deprecated `managed_roles`), this means a declared view/matview GRANT is re-emitted on every run and **never converges** — dry-run never reaches zero diff. `objectOwners()` already covers views/matviews (`relkind in ('r','p','v','m')`), so declarative OWNER management works, while GRANT management silently drops the same objects. No REVOKE is produced, so there is no data loss — but view privileges cannot be managed declaratively.

## Reproduction

```sql
CREATE SCHEMA app AUTHORIZATION app_owner;
CREATE TABLE app.t (id int PRIMARY KEY); ALTER TABLE app.t OWNER TO app_owner;
CREATE VIEW app.v_t AS SELECT id FROM app.t; ALTER VIEW app.v_t OWNER TO app_owner;
GRANT SELECT ON app.v_t TO app_reader;   -- grant on a view
```

```yaml
# config.yml
manage:
  privilege:
    - target: 'app_owner|app_reader'
      drop: true
```

```sql
-- desired.sql declares the same view grant
CREATE TABLE "app"."t" (...);
ALTER TABLE app.t OWNER TO app_owner;
CREATE VIEW "app"."v_t" AS SELECT id FROM app.t;
ALTER TABLE app.v_t OWNER TO app_owner;
GRANT SELECT ON TABLE app.v_t TO app_reader;
```

```
$ psqldef --config config.yml --dry-run <db> < desired.sql
GRANT SELECT ON TABLE "app"."v_t" TO "app_reader";   # re-emitted every run — never converges
```

The export does not see the existing view grant, so the declared grant always looks missing.

## Fix

Add `viewPrivileges()`, which reads `relacl` (table-level) and `attacl` (column-level) for `relkind in ('v','m')` and emits standalone `GRANT ... ON TABLE` statements, mirroring `sequencePrivileges()` and `objectOwners()`. `information_schema.table_privileges` does not expose materialized views, so `relacl` is read directly to cover `'v'` and `'m'` uniformly. The same `TargetSchema` and managed-grantee filters as the other export helpers are applied. Only runs when privilege management is configured, so `--export` output is unchanged for users who do not manage privileges.

## Test

Adds view/matview GRANT cases to `tests_manage_privilege.yml`:
- view GRANT idempotent convergence (guards the bug via the current-schema idempotency check),
- view GRANT drift → `REVOKE` (with `drop: true`),
- matview GRANT idempotent convergence (guards the relacl-for-matview path).

Verified that the new tests fail without the fix (view/matview grants are re-emitted and never converge) and pass with it. Full `cmd/psqldef` + `database/postgres` suite green locally on PG18.

### DIFF
--- a/cmd/psqldef/tests_manage_privilege.yml
+++ b/cmd/psqldef/tests_manage_privilege.yml
@@ -276,3 +276,77 @@ ManagePrivilegeViewOwner:
   up: |
     ALTER TABLE v_users OWNER TO app_user;
   down: ""
+
+ManagePrivilegeViewGrantIdempotent:
+  # A GRANT on a view must be seen by the export so a declared view grant
+  # converges. Without view privilege export the current schema is not
+  # idempotent (the grant is re-emitted every run).
+  manage:
+    privilege:
+      - target: app_user
+        drop: true
+  current: |
+    CREATE TABLE users (
+      id BIGINT PRIMARY KEY,
+      name VARCHAR(100)
+    );
+    CREATE VIEW v_users AS SELECT id FROM users;
+    GRANT SELECT ON TABLE v_users TO app_user;
+  desired: |
+    CREATE TABLE users (
+      id BIGINT PRIMARY KEY,
+      name VARCHAR(100)
+    );
+    CREATE VIEW v_users AS SELECT id FROM users;
+    GRANT SELECT ON TABLE v_users TO app_user;
+  up: ""
+  down: ""
+
+ManagePrivilegeViewGrantRevoke:
+  # Dropping a declared view grant yields a REVOKE when the rule allows drop.
+  manage:
+    privilege:
+      - target: app_user
+        drop: true
+  current: |
+    CREATE TABLE users (
+      id BIGINT PRIMARY KEY,
+      name VARCHAR(100)
+    );
+    CREATE VIEW v_users AS SELECT id FROM users;
+    GRANT SELECT ON TABLE v_users TO app_user;
+  desired: |
+    CREATE TABLE users (
+      id BIGINT PRIMARY KEY,
+      name VARCHAR(100)
+    );
+    CREATE VIEW v_users AS SELECT id FROM users;
+  up: |
+    REVOKE SELECT ON TABLE "public"."v_users" FROM "app_user";
+  down: |
+    GRANT SELECT ON TABLE "public"."v_users" TO "app_user";
+
+ManagePrivilegeMatViewGrantIdempotent:
+  # information_schema.table_privileges does not expose materialized views, so
+  # view privilege export reads relacl directly. This guards that matview
+  # grants converge too.
+  manage:
+    privilege:
+      - target: app_user
+        drop: true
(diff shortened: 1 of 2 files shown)


moznion (commented):
Thank you for your contribution.

The overall direction of this pull request looks good. However, this change causes dropping a VIEW with GRANT statements to fail because the VIEW’s GRANT statements are still exported. As a result, the GRANT entries remain in `currentPrivileges` even after the VIEW has been dropped.

For example, the following test case would be failed:

```yaml
ManagePrivilegeDropViewWithGrant:
  legacy_ignore_quotes: false
  manage:
    privilege:
      - target: app_user
        drop: true
  current: |
    CREATE TABLE users (
      id bigint PRIMARY KEY,
      name varchar(100)
    );
    CREATE VIEW v_users AS SELECT id FROM users;
    GRANT SELECT ON TABLE v_users TO app_user;
  desired: |
    CREATE TABLE users (
      id bigint PRIMARY KEY,
      name varchar(100)
    );
  up: |
    DROP VIEW public.v_users;
  down: |
    CREATE VIEW public.v_users AS SELECT id FROM users;
    GRANT SELECT ON TABLE public.v_users TO app_user;
```

Do you think you could address this issue?

dip-daiki-ogikubo:
Thanks for reviewing, and for the test case — it found a real bug, though not one this PR introduced.

**This PR is now redundant.** All four cases it added (view GRANT idempotent, view GRANT revoke, column-level view GRANT, matview GRANT) pass on master as of 8bb8c6e: #1376 exports view and materialized view privileges, which is what this PR set out to add. Rebasing it onto master leaves only a duplicate `REVOKE`, since the grants are then exported twice. Closing it.

**Your `ManagePrivilegeDropViewWithGrant` case reproduces on master without this PR:**

```
DROP VIEW public.v_users;
REVOKE SELECT ON TABLE public.v_users FROM "app_user";
pq: relation "public.v_users" does not exist (42P01)
```

On e162fd7 (the commit before #1376) the same test fails the other way — the view grant is not exported at all, so the current schema is not idempotent — so this is a gap that opened when view privileges started being exported.

The cause is that `isPrivilegeOnRemovedTable` only consults `droppedTableNames`, which is filled when a table is dropped; the loop that drops views records nothing, so a dropped view's privileges are still treated as orphaned. Fix and tests for views and matviews are in #1379.

dip-daiki-ogikubo:
Superseded by #1376 (which covers the view/matview GRANT export this PR added) and #1379 (which fixes the DROP VIEW + REVOKE ordering found in review).

PR #1335, opened 2026-08-20
psqldef: support `FOR EACH STATEMENT` triggers

moznion:
The generic parser only accepted `FOR EACH ROW` in `CREATE TRIGGER`, so statement-level triggers failed to parse, both in desired schemas and in `pg_get_triggerdef()` output on export. The generator also hardcoded `FOR EACH ROW` for PostgreSQL, so a trigger written without a `FOR EACH` clause (which PostgreSQL defines as defaulting to `STATEMENT`; ref: https://www.postgresql.org/docs/16/sql-createtrigger.html) was silently created as a row-level trigger.

- Accept `FOR EACH STATEMENT` in the PostgreSQL `CREATE TRIGGER` rules (with and without `WHEN`, `EXECUTE FUNCTION`/`PROCEDURE`) and record the `FOR EACH` scope in the AST
- Generate `FOR EACH <scope>`, resolving an omitted clause to `STATEMENT` per PostgreSQL semantics, and compare the scope in areSameTriggerDefinition with the same normalization
- Add `STATEMENT` as a non-reserved keyword; goyacc conflict counts are unchanged

---

EDIT:
This would be a breaking change because the current version of psqldef generates `FOR EACH ROW` if the `FOR EACH` clause is omitted, while the new version generates `FOR EACH STATEMENT`. Do we need to add any safeguards to preserve backward compatibility? e.g., prohibiting omission of the FOR EACH clause.


moznion:
@gfx I don’t mean to rush you, but this is currently blocking our project. Could you please take a look when you have a chance?

PR #1337, opened 2026-08-24
psqldef: skip REVOKE for privileges on dropped or renamed tables

draftcode:
## Problem

With `managed_roles` (or `manage.privilege`) configured, dropping a table produces DDLs that cannot be applied:

```sql
DROP TABLE "public"."audit_logs";
REVOKE SELECT, INSERT ON TABLE "public"."audit_logs" FROM "app_user";
```

The `REVOKE` fails with `42P01 relation does not exist`. Inside a transaction the `DROP` is already visible to the following statement, so the whole apply rolls back; with `--disable-ddl-transaction` (or on an engine without DDL transactions) the `DROP` commits and the run aborts halfway.

Reproduction:

```sql
-- current
CREATE TABLE audit_logs (id BIGINT PRIMARY KEY);
GRANT SELECT, INSERT ON TABLE audit_logs TO app_user;

-- desired: table removed
```

```sh
psqldef --enable-drop --config config.yml   # managed_roles: [app_user]
```

The same happens for a table renamed through `-- @renamed from=`, where the `REVOKE` names the table's old name:

```sql
ALTER TABLE "public"."old_logs" RENAME TO "new_logs";
GRANT SELECT ON TABLE "public"."new_logs" TO "app_user";
REVOKE SELECT ON TABLE "public"."old_logs" FROM "app_user";
```

`--enable-drop` is only a gate on whether destructive statements may be emitted, so it cannot be used to work around this.

## Cause

When a table is dropped, `generateDDLs` prunes `currentTables` and records `droppedTables`, but `currentPrivileges` is left untouched. The orphaned-privilege cleanup at the end of `generateDDLs` then finds the dropped table's GRANT with no counterpart in the desired schema, concludes the grantee is orphaned, and emits a REVOKE — appended after the DROP TABLE that the same run already emitted.

## Fix

Neither REVOKE is needed in the first place: PostgreSQL removes privileges together with the table, and carries them over on a rename. So this skips the orphan check for tables dropped or renamed in the same run, rather than reordering the statements.

Dropped and renamed table names are tracked as `QualifiedName` alongside the existing `droppedTables` map, because the privilege's table name has to be compared through the default-schema normalization in `qualifiedNamesEqual`, which a raw-string key lookup would miss. This mirrors how COMMENT cleanup is already skipped for dropped tables via `isCommentOnDroppedTable`.

The drop side of the check is gated on `enable_drop`: without it the DROP TABLE is only commented out, so the table survives and its privileges are still revocable. That combination reaches an executable REVOKE only under `manage.privilege`, which keeps REVOKEs out of the `enable_drop` commenting pass, and that REVOKE stays as it is today. A rename is not gated, so it is always skipped.

## Tests

Added `ManagedRolesDropTable` and `ManagedRolesRenameTable` to `cmd/psqldef/tests_managed_roles.yml`. Both fail on master with the stray REVOKE and pass with this change. `ManagePrivilegeDropTableWithoutEnableDrop` in `cmd/psqldef/tests_manage_privilege.yml` pins the `enable_drop: false` behavior that is deliberately left unchanged.

Verified against PostgreSQL 18 with `PSQLDEF_PARSER=generic`: `./cmd/psqldef`, `./schema`, `./parser`, `./database/postgres` and `./cmd/sqlite3def` all pass.


moznion (approved):
Thank you for your contribution. Looks great.

PR #1338, opened 2026-08-24
psqldef: apply `--skip-extension` to all schema inputs

asfrgrtgd:
## What

Make `--skip-extension` an absolute opt-out for extensions across every schema source handled by psqldef.

- Add `SkipExtension` to `database.GeneratorConfig` and propagate the CLI flag after merging `--config` and `--config-inline`.
- Update the shared `schema.FilterExtensions` path to remove every extension before evaluating `manage.extension`. Because the generator applies this filter to both desired and current DDLs, the flag now works for live comparisons, offline comparisons using a current SQL file, and file-based export.
- Keep the PostgreSQL exporter's existing early skip, avoiding the extension catalog query for live databases.
- Document the flag's scope and its precedence over `manage.extension`.

When `--skip-extension` and `manage.extension` are both specified, `--skip-extension` wins: psqldef does not create, drop, or export any extension. Without the flag, existing extension management and `manage.extension` behavior remain unchanged.

## Motivation

Previously, `--skip-extension` was passed only to the database exporter. The PostgreSQL live exporter used it to omit extensions from the current schema, but the shared generator never received the flag. Extensions from the desired schema or an offline current SQL file therefore remained in the diff.

This produced inconsistent behavior depending on the schema source:

| Scenario | Before | After |
|---|---|---|
| Extension exists only in the desired schema | Emit `CREATE EXTENSION` | No DDL |
| Extension exists only in an offline current SQL file | Emit `DROP EXTENSION` | No DDL |
| `--skip-extension` is combined with matching `manage.extension` rules | Rules could keep extensions in the diff | No extension is managed |
| Live current schema or live `--export` | Extensions were already skipped by the exporter | Behavior preserved |

Applying the flag in the shared filter gives current and desired schemas the same semantics regardless of whether the current schema comes from a live database or a file.

## Test

- `cmd/psqldef/psqldef_test.go` — expand `TestPsqldefSkipExtension` to cover live current schema, desired schema, offline current SQL, export, and precedence over `manage.extension`.
- `schema/generator_test.go` — add DB-less coverage for desired-only and current-only extensions, `manage.extension` precedence, and preservation of non-extension DDLs.

### DIFF
--- a/cmd-psqldef.md
+++ b/cmd-psqldef.md
@@ -546,6 +546,8 @@ $ psqldef -U postgres dbname --apply \
 
 By default, psqldef manages every extension in the database (equivalent to `--skip-extension` being off). Some managed PostgreSQL services (e.g. AlloyDB) auto-install extensions such as `google_columnar_engine` that should never be diffed or dropped, while `--skip-extension` would also stop managing extensions you do want tracked, like `vector` or `pg_trgm`.
 
+`--skip-extension` excludes all extensions from both the current and desired schemas. It applies consistently to live database comparisons, offline comparisons using a current SQL file, and `--export`. The flag takes precedence over `manage.extension`, so no extension is created, dropped, or exported when both are specified. To manage only selected extensions, omit `--skip-extension` and use `manage.extension`.
+
 `manage.extension` restricts management to extensions matching a rule, leaving everything else untouched:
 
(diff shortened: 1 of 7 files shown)


moznion (approved):
Thank you for your contribution, looks nice.

PR #1339, opened 2026-08-25
psqldef: preserve time zone modifier in timestamp/time cast targets

dip-daiki-ogikubo:
## Problem

Regenerating a column `DEFAULT` drops the timezone modifier from a cast in the default expression, so `... with time zone` becomes `... without time zone` — a **different type**. The applied default then differs from the declaration, and the next dry-run reports no diff (silent). `ADD COLUMN` is affected the same way (it shares the default serialization), producing the column type right and the default wrong.

- psqldef v3.11.20 (`29164ef`), PostgreSQL 16

## Reproduction

```sql
CREATE SCHEMA s;
CREATE TABLE s.t (a timestamp(0) with time zone NOT NULL
       DEFAULT (clock_timestamp())::timestamp(0) with time zone);
```

Change only the default (keep the modifier):

```sql
CREATE TABLE "s"."t" (
    "a" timestamp(0) with time zone NOT NULL DEFAULT (now())::timestamp(0) with time zone
);
```

```
$ psqldef --dry-run ... < desired.sql
ALTER TABLE "s"."t" ALTER COLUMN "a" SET DEFAULT now()::timestamp(0);   # with time zone dropped
```

Re-running produces no diff even though the stored default no longer matches the declaration.

## Root cause

The modifier survives the parse — the pgquery converter keeps it in the type name (`timestamptz`), and `getRawTypeName` returns it unchanged. It is lost one layer later in `normalizeConvertType` (`schema/normalize.go`): the `normalizeTypeName` alias map folds `timestamptz`/`timetz` (and the spelled-out `... with time zone` forms) down to `timestamp`/`time`. That is correct for `ColumnType`, which carries a separate `Timezone` flag, but lossy for `ConvertType`, which has no such flag. Because `generateDefaultDefinition` emits the comparison-canonical (normalized) form as the actual SQL, this lossy normalization rewrites the cast's type rather than being a mere comparison artifact.

There is also a grammar gap: `simple_convert_type` accepts `TIMESTAMP '(' INTEGRAL ')'` but not a following `WITH TIME ZONE`, so `(now())::timestamp(0) with time zone` fails the generic parser and the whole DDL falls back to pgquery (the `Generic parser failed ... using pgquery fallback` warnings).

## Fix

- **`normalize.go` (load-bearing):** in `normalizeConvertType`, recover the timezone modifier from the type-name aliases before `normalizeTypeName` erases it. This fixes every spelling on the path that runs — `::timestamptz`, `::timestamp with time zone` (with or without precision), `CAST(... AS ...)`, and the case where an unrelated non-generic-parseable statement forces the whole file onto the pgquery path.
- **`parser.y` / `node.go` (secondary):** accept `time_zone_opt` after the precision on `TIMESTAMP`/`TIME` cast targets and add a `TimeZone` field to `ConvertType`. This removes the fallback warning and lets the cast round-trip through the generic parser. On its own it only covers the exact `timestamp(p) with time zone` spelling, which is why the normalize fix is the primary one.

## Tests

- `schema`: unit test asserting `normalizeConvertType` recovers the modifier for `timestamptz` / `timetz` / spelled-out forms (fails without the fix).
- `parser`: parse + round-trip guards for `::timestamp(p) with time zone` and `::time(p) with time zone`.
- `cmd/psqldef`: `SET DEFAULT` change/idempotency and `ADD COLUMN` cases keeping `with time zone`.

Full `parser`, `schema`, `database/postgres`, and `cmd/psqldef` (`PG_FLAVOR=pgvector`) suites pass locally on PG18; goyacc conflict counts unchanged (28 shift/reduce, 545 reduce/reduce).

## Note on the impact framing

The stored values happen to match in the exact repro because a `timestamptz` column coerces the naive value back — but only in a zone with no DST. In a DST zone the naive value is ambiguous for one hour a year and does not round-trip; and for a non-timestamptz target (e.g. a `text` column with `DEFAULT (now())::timestamptz(0)` vs `::timestamp(0)`) the offset is genuinely lost. Same invisible diff, real data change. (Thanks to @ChangkeunJ for pinpointing the normalize layer and these sharper cases.)


moznion (commented):
Thank you for your work.

`normalizeConvertType` handles arrays via `TrimSuffix(typeStr, "[]")`, but `ConvertType.Format` emits `TimeZone` after the `[]`, so array casts produce invalid SQL. Where master emits the wrong type but still applies, this PR fails to apply at all:

```
# this PR
... SET DEFAULT '{2020-01-01}'::timestamp[] with time zone;
pq: syntax error at or near "with" at column 83 (42601)
```

```
# master (29164ef)
... SET DEFAULT '{2020-01-01}'::timestamp[];   # applies
```

`CreateTableWithDefault` in `tests_datatypes.yml` already has `'{}'::int[]` and `ARRAY[]::int[]`, so this is a reachable path.

An example of failing test case:


```yaml
AddColumnDefaultTimestampWithTimeZoneArrayCast:
  current: |
    CREATE TABLE test (
      id integer
    );
  desired: |
    CREATE TABLE test (
      id integer,
      a timestamptz[] DEFAULT '{}'::timestamptz[]
    );
  up: |
    ALTER TABLE public.test ADD COLUMN a timestamp WITH TIME ZONE[] DEFAULT '{}'::timestamp with time zone[];
  down: |
    ALTER TABLE public.test DROP COLUMN a;
  legacy_ignore_quotes: false
```

What do you think about this behavior? Should this issue be addressed in this pull request, or should it be handled separately?

dip-daiki-ogikubo:
Good catch, and thank you for the falsifiable case — that is a real regression, so I fixed it here in c670242 rather than deferring it.

The `[]` suffix lives in `ConvertType.Type`, while `Format` appended the length and the time zone modifier after the whole type name, so an array cast came out as `timestamp[] with time zone`. `Format` now splits the suffix off and re-appends it after both, which also fixes the length placement that had the same problem (`timestamp[](0)` -> `timestamp(0)[]`).

* `AddColumnDefaultTimestampWithTimeZoneArrayCast` in `tests_datatypes.yml` is your case, and it now produces exactly the DDL you wrote.
* Array cases were added to the normalize round-trip test (`timestamptz[]` with and without precision).
* Both fail without the change (`'{}'::timestamp[] with time zone`).

Also rebased onto master: `parser.go` was regenerated (conflicts 28 shift/reduce, 286 reduce/reduce — identical to master's own grammar), and the expected round-trip strings now carry the parentheses that #1361 preserves. Full `parser`, `schema`, `database`, `database/postgres`, `cmd/psqldef` and `cmd/sqlite3def` suites pass on PG18.

moznion (approved):
This seems nice. Thank you for your contribution.

PR #1341, opened 2026-08-25
psqldef: diff split desired GRANTs for the same object/grantee once

dip-daiki-ogikubo:
## Problem

When the privileges for the same `(object, grantee, grant option)` are declared across **several** desired `GRANT` statements, psqldef emits the resulting `REVOKE` (or `GRANT`) **once per source statement** — identical duplicated lines — under `manage.privilege` with `drop: true`.

- psqldef: v3.11.20 (`29164ef`)
- PostgreSQL: 16 (`postgres:16` image)

## Reproduction

```sql
CREATE SCHEMA dip_owner;
CREATE ROLE app_x NOLOGIN;
CREATE TABLE dip_owner.t1 (id int);
GRANT SELECT, INSERT, UPDATE, DELETE ON dip_owner.t1 TO app_x;
```

```yaml
# config.yml
target_schema: dip_owner
manage:
  privilege:
    - target: 'app_x'
      drop: true
```

**One desired statement (correct):**

```sql
GRANT SELECT ON TABLE dip_owner.t1 TO app_x;
```
```
REVOKE DELETE, INSERT, UPDATE ON TABLE "dip_owner"."t1" FROM "app_x";
```

**Split into two desired statements (bug):**

```sql
GRANT SELECT ON TABLE dip_owner.t1 TO app_x;
GRANT INSERT ON TABLE dip_owner.t1 TO app_x;
```
```
REVOKE DELETE, UPDATE ON TABLE "dip_owner"."t1" FROM "app_x";
REVOKE DELETE, UPDATE ON TABLE "dip_owner"."t1" FROM "app_x";   # duplicated
```

The GRANT side duplicates the same way (e.g. current has only `SELECT`, desired declares `SELECT` and `INSERT` as two statements → `GRANT INSERT` is emitted twice).

Splitting grants across statements happens naturally when the desired schema is generated from the catalog: `aclexplode(pg_class.relacl)` yields one row per privilege.

## Why it matters

The DDL is idempotent, so no data-level harm, but:

1. **Plan readability** — reviewers see identical lines repeated.
2. **Row-count safety checks** — a caller that blocks an apply when the number of REVOKEs exceeds a threshold counts one logical revoke as N.
3. **Audit records** carry the duplicates.

## Root cause

Desired GRANTs for the same `(object, grantees, grant option, object type)` are already merged into a single entry in `g.desiredPrivileges` (`convertDDLsToTablesAndViews`), and `generateDDLsForGrantPrivilege` already cross-references that aggregated set to compute the correct REVOKE/GRANT. But the generation loop iterates the **raw** desired DDLs and calls `generateDDLsForGrantPrivilege` once per raw `GrantPrivilege`, so the same diff is produced repeatedly.

## Fix

Route each raw desired `GrantPrivilege` to its aggregated entry and generate the diff **once per aggregated entry** (tracked by identity). This diffs the full privilege set as one unit, so a split declaration produces a single `REVOKE`/`GRANT`. Simply skipping duplicate statements would be wrong — it would drop privileges that only a later statement declares — which is why the aggregated (merged) entry is used.

## Tests

Adds `tests_manage_privilege.yml` cases for split-statement `REVOKE` and `GRANT`. Verified they fail without the fix (duplicated lines) and pass with it. Full `parser`, `schema`, `database/postgres`, and `cmd/psqldef` (`PG_FLAVOR=pgvector`) suites pass locally on PG18.

### DIFF
--- a/schema/generator.go
+++ b/schema/generator.go
@@ -235,6 +235,12 @@ func (g *Generator) generateDDLs(desiredDDLs []DDL) ([]string, error) {
 	// bulkAlter fuses per-table ALTER TABLE actions when --bulk-alter is set (MySQL only).
 	bulkAlter := newAlterBundler(g, g.config.BulkAlter && g.mode == GeneratorModeMysql)
 
+	// Desired GRANTs for the same (object, grantees, grant option) can be split
+	// across several statements; they are merged into one entry in
+	// g.desiredPrivileges. Track which aggregated entry has already produced its
+	// diff so the REVOKE/GRANT is emitted once, not once per source statement.
+	generatedGrantPrivileges := map[*GrantPrivilege]bool{}
+
 	// Incrementally examine desiredDDLs
 	for _, ddl := range desiredDDLs {
 		switch desired := ddl.(type) {
@@ -411,7 +417,19 @@ func (g *Generator) generateDDLs(desiredDDLs []DDL) ([]string, error) {
 			}
 			createSchemaDDLs = append(createSchemaDDLs, schemaDDLs...)
 		case *GrantPrivilege:
-			privilegeDDLs, err := g.generateDDLsForGrantPrivilege(desired)
+			// Diff the merged desired privileges for this (object, grantees,
+			// grant option) once. Using the aggregated entry keeps split GRANT
+			// statements from each re-emitting the same REVOKE/GRANT, while still
+			// covering every privilege they collectively declare.
+			aggregated := g.aggregatedDesiredPrivilege(desired)
+			if aggregated == nil {
+				aggregated = desired
+			}
+			if generatedGrantPrivileges[aggregated] {
+				break
+			}
+			generatedGrantPrivileges[aggregated] = true
+			privilegeDDLs, err := g.generateDDLsForGrantPrivilege(aggregated)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


schema/generator.go
```diff
@@ -411,7 +417,19 @@ func (g *Generator) generateDDLs(desiredDDLs []DDL) ([]string, error) {
 			}
 			createSchemaDDLs = append(createSchemaDDLs, schemaDDLs...)
 		case *GrantPrivilege:
-			privilegeDDLs, err := g.generateDDLsForGrantPrivilege(desired)
+			// Diff the merged desired privileges for this (object, grantees,
+			// grant option) once. Using the aggregated entry keeps split GRANT
+			// statements from each re-emitting the same REVOKE/GRANT, while still
+			// covering every privilege they collectively declare.
+			aggregated := g.aggregatedDesiredPrivilege(desired)
```
moznion, line 424: I guess it might be able to determine whether it is the aggregation target simply by checking for pointer equality, whether that raw statement is contained in `g.desiredPrivileges`.

e.g.,

```go
case *GrantPrivilege:
    if !slices.Contains(g.desiredPrivileges, desired) {
        break
    }
    ...
```

Accordingly, `aggregatedDesiredPrivilege` would be removable.

moznion (commented):
Thank you for your contribution. I left just one comment.

schema/generator.go
dip-daiki-ogikubo: Thanks — done in a22e4a1.

You are right that membership is the whole signal: `aggregateDDLsToSchema` keeps the first statement of each merged group in `g.desiredPrivileges` and folds the other statements' privileges into that entry, so a raw GRANT is either that entry or was merged away. `slices.Contains(g.desiredPrivileges, desired)` replaces both the field-by-field lookup and the map that tracked which entry had been generated, and `aggregatedDesiredPrivilege` is gone (-34/+7).

Rebased on master. `parser`, `schema`, `database`, `database/postgres` and `cmd/psqldef` all pass on PG18, and removing the check brings the three split-GRANT tests back to failing.

moznion:
Thank you for your work, LGTM

PR #1342, opened 2026-08-25
psqldef: make IS NOT DISTINCT FROM idempotent in CHECK constraints and partial indexes

178inaba:
## Why

PostgreSQL rewrites `a IS NOT DISTINCT FROM b` into `NOT (a IS DISTINCT FROM b)` when it
stores an expression. psqldef compared the desired expression against the stored one
without folding the two spellings together, so a schema written with `IS NOT DISTINCT
FROM` never converged:

- **CHECK constraint**: `DROP CONSTRAINT` + `ADD CONSTRAINT` on every run. `ADD CONSTRAINT
  ... CHECK` revalidates every existing row, so this takes a real lock on a large table
  each time.
- **Partial index**: without `--enable-drop` the `DROP INDEX` is skipped while the `CREATE
  INDEX` is still emitted, so psqldef exits 1 with `42P07 relation "..." already exists`
  and the whole transaction rolls back — taking any unrelated DDL in the same run with it.

Reproduced on v3.11.20 against PostgreSQL 18, with the same output on every run:

```console
$ psqldef ... --file schema.sql repro
-- Apply --
BEGIN;
ALTER TABLE "public"."products" DROP CONSTRAINT "products_status_check";
ALTER TABLE "public"."products" ADD CHECK (status is not distinct from 'pending');
-- Skipped: DROP INDEX "public"."idx_orders_pending";
CREATE INDEX idx_orders_pending ON orders (id) WHERE status IS NOT DISTINCT FROM 'pending';
ROLLBACK;
pq: relation "idx_orders_pending" already exists (42P07)
exit status 1
```

This is the same class of bug as #1295 (`IN (...)` in CHECK constraints and partial
indexes) and #1182, for a different operator. The parser already builds the right AST for
both spellings; only the comparison path was missing the fold.

## What

- `normalizeNotExpr` folds `NOT (a IS DISTINCT FROM b)` into a single `IS NOT DISTINCT
  FROM` comparison. Both normalizers call it, the way they already share
  `normalizeComparisonExpr` — #1295 had to fix `IN (...)` twice because the two had
  drifted apart.
- `normalizeExpr` had no `NotExpr` case at all: a `NOT (...)` predicate fell through to
  `default:` and was compared with nothing inside it normalized. The new case normalizes
  the operand like every other case does, which also fixes cast and quoting normalization
  inside `NOT (...)` predicates.

### Why the fold is one-way

Measured on PostgreSQL 18 with `pg_get_constraintdef` / `pg_get_expr(indpred, indrelid)`:

| written | stored |
|---|---|
| `a IS DISTINCT FROM b` | `a IS DISTINCT FROM b` (unchanged) |
| `a IS NOT DISTINCT FROM b` | `NOT (a IS DISTINCT FROM b)` |
| `NOT (a IS NOT DISTINCT FROM b)` | `NOT (NOT (a IS DISTINCT FROM b))` |

PostgreSQL rewrites the operator but does not collapse the resulting double negation, so
folding only the inner `NOT (... IS DISTINCT FROM ...)` already lands both sides on the
same shape. No reverse fold and no double-negation elimination is needed.

Canonicalizing towards `IS NOT DISTINCT FROM` keeps the generated DDL in the spelling the
author wrote. Both formatters print `ComparisonExpr.Operator` verbatim, so unlike #1295 —
where `formatExprQuoteAware` dropped the `Any`/`All` flags — no formatter change is needed.

### Why the parentheses are restored

`normalizeExpr` strips `ParenExpr` for PostgreSQL and MySQL, and `NotExpr` prints without
parentheses. Recursing into the operand would therefore turn `NOT (a AND b)` into `not a
and b`, which is string-equal to the very different `NOT a AND b` (these comparisons are
string, not AST, comparisons) and would also drop the parentheses from the DDL psqldef
emits for `DEFAULT` expressions. `normalizeNotExpr` puts them back around `AND` / `OR`
operands, which are the only nodes the grammar binds looser than `NOT`.
`PartialIndexWithWhereNotAndPrecedence` pins this: with the branch removed it fails with an
empty diff where a `DROP INDEX` + `CREATE INDEX` is expected.

### Not gated on PostgreSQL

`NOT (a IS DISTINCT FROM b)` ≡ `a IS NOT DISTINCT FROM b` in any dialect, so the fold runs
in every `GeneratorMode` rather than adding a branch. SQL Server 2025 stores the operator
verbatim (`([a] IS NOT DISTINCT FROM [b])`), so the fold is a no-op there, and MySQL and
SQLite have no such operator.

## Tests

New psqldef YAML cases cover both spellings, the double negation, and the already working
`IS DISTINCT FROM` direction, in `tests_constraints.yml` (including one legacy-quotes case,
since CHECK output goes through `parser.String` there and `formatExprQuoteAware` otherwise)
and `tests_indices.yml`.

The tests are psqldef-only on purpose: SQL Server stores the operator verbatim and
MySQL/SQLite do not have it, so cases there would pass without this change and assert
nothing.

`make build`, `make test-all-flavors`, `make format` and `make lint` all pass; the psqldef
suite was also run with `PSQLDEF_PARSER=generic`.


cmd/psqldef/tests_constraints.yml
```diff
@@ -1095,6 +1095,54 @@ ConstraintCheckArrayColumnReferenceMixedTypesQuoteAware:
...
+    );
+  desired: |
+    CREATE TABLE products (
+      id serial,
+      status text CHECK (NOT (status IS DISTINCT FROM 'pending'))
+    );
+  up: |
+    ALTER TABLE public.products ADD CHECK (status is not distinct from 'pending');
+  down: |
+    ALTER TABLE public.products DROP CONSTRAINT products_status_check;
```
moznion: This pull request appears to fix the semantics of CHECK constraints. For example, it parses `CHECK (NOT a BETWEEN 1 AND 2)` as:

base: `CHECK (not a >= 1 and a <= 2)` (i.e., `(NOT a>=1) AND a<=2`)
this PR: `CHECK (not (a >= 1 and a <= 2))`

The change introduced by this pull request looks correct, but there doesn’t seem to be a test that verifies this behavior. Could you please add a test case for it?

moznion (approved):
Thank you!

PR #1344, opened 2026-08-30
psqldef: compare EXTRACT and date_part according to the PostgreSQL server version

asfrgrtgd:
## What

psqldef compares `EXTRACT` and `date_part` according to the PostgreSQL server version.

- PostgreSQL 13 and earlier normalize built-in `date_part` calls to `EXTRACT` across the VIEW AST, including window expressions and `WITHIN GROUP`.
- PostgreSQL 14 and later preserve the type-sensitive distinction between `EXTRACT` and `date_part`.
- VIEW and MATERIALIZED VIEW export preserves the definition rendered by PostgreSQL.

## Motivation

On PostgreSQL 13 and earlier, exported VIEW definitions can render `EXTRACT` as `date_part`. Comparing the two forms literally recreates an unchanged VIEW. On PostgreSQL 14 and later, an unconditional conversion in the exporter has the opposite problem: it treats expressions with different return types as equal and can hide a VIEW change.

psqldef derives the comparison rule from `server_version_num` and applies it during AST normalization. PostgreSQL 13 and earlier canonicalize only built-in `date_part` calls, including nested and schema-qualified forms. PostgreSQL 14 and later keep `EXTRACT` and `date_part` distinct. The exporter retains PostgreSQL's rendered definition.

## Test

- `database/postgres/database_test.go`
  - Verify the exact `139999` / `140000` version boundary.
  - Verify conservative comparison when version lookup fails.
  - Verify that MATERIALIZED VIEW export preserves `date_part`.
- `schema/generator_test.go`
  - Verify built-in, case-folded, quoted, and other schema-qualified `date_part` forms.
  - Verify normalization in window expressions and `WITHIN GROUP`.
  - Verify generated VIEW DDL with equivalence enabled and disabled.
- `cmd/psqldef/tests_views.yml`
  - Verify zero VIEW DDL on PostgreSQL 13 and earlier.
  - Verify zero VIEW DDL for window and ordered-set aggregate expressions on PostgreSQL 13 and earlier.
  - Verify bidirectional VIEW recreation on PostgreSQL 14 and later.

### DIFF
--- a/cmd/psqldef/tests_views.yml
+++ b/cmd/psqldef/tests_views.yml
@@ -493,6 +493,87 @@ ViewWithGroupByAndHaving:
     CREATE VIEW tag_counts AS
     SELECT tag, COUNT(*) AS count
     FROM taggings GROUP BY tag HAVING COUNT(*) > 1;
+ExtractAndDatePartEquivalentBeforePostgres14:
+  max_version: "13"
+  legacy_ignore_quotes: false
+  current: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_years AS
+      SELECT
+        EXTRACT(year FROM created_at) AS event_year,
+        EXTRACT(month FROM created_at) AS event_month
+      FROM events;
+  desired: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_years AS
+      SELECT
+        date_part('year', created_at) AS event_year,
+        pg_catalog.date_part('month', created_at) AS event_month
+      FROM events;
+  up: ""
+  down: ""
+ExtractAndDatePartEquivalentInWindowAndWithinGroupBeforePostgres14:
+  max_version: "13"
+  legacy_ignore_quotes: false
+  current: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_windows AS
+      SELECT row_number() OVER (
+        PARTITION BY EXTRACT(month FROM created_at)
+        ORDER BY EXTRACT(year FROM created_at)
+      ) AS row_num
+      FROM events;
+    CREATE VIEW event_percentiles AS
+      SELECT percentile_cont(0.5) WITHIN GROUP (
+        ORDER BY EXTRACT(epoch FROM created_at)
+      ) AS median_epoch
+      FROM events;
+  desired: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_windows AS
+      SELECT row_number() OVER (
+        PARTITION BY date_part('month', created_at)
+        ORDER BY date_part('year', created_at)
+      ) AS row_num
+      FROM events;
+    CREATE VIEW event_percentiles AS
+      SELECT percentile_cont(0.5) WITHIN GROUP (
+        ORDER BY date_part('epoch', created_at)
+      ) AS median_epoch
+      FROM events;
+  up: ""
+  down: ""
+ExtractAndDatePartDifferFromPostgres14:
+  min_version: "14"
+  legacy_ignore_quotes: false
+  current: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_years AS
+      SELECT EXTRACT(year FROM created_at) AS event_year
+      FROM events;
+  desired: |
+    CREATE TABLE events (
+      created_at timestamp
+    );
+    CREATE VIEW event_years AS
+      SELECT date_part('year', created_at) AS event_year
+      FROM events;
+  up: |
+    DROP VIEW public.event_years;
+    CREATE VIEW public.event_years AS select date_part('year', created_at) as event_year from events;
+  down: |
+    DROP VIEW public.event_years;
+    CREATE VIEW public.event_years AS select EXTRACT(year FROM created_at) as event_year from events;
 ExtractInMaterializedView:
   desired: |
     CREATE TABLE events (
--- a/database/postgres/database.go
+++ b/database/postgres/database.go
@@ -56,10 +56,26 @@ func (d *PostgresDatabase) SetGeneratorConfig(config database.GeneratorConfig) {
 		d.defaultOpclasses = map[string]bool{}
 	}
 	config.PostgresDefaultOperatorClasses = d.defaultOpclasses
+	config.PostgresExtractDatePartEquivalent = false
+	var serverVersionNum int
+	if err := d.db.QueryRow("SHOW server_version_num").Scan(&serverVersionNum); err != nil {
+		slog.Debug("Failed to get PostgreSQL server_version_num", "error", err)
+	} else {
+		config.PostgresExtractDatePartEquivalent = extractDatePartEquivalent(serverVersionNum)
+		slog.Debug(
+			"Determined PostgreSQL EXTRACT/date_part comparison capability",
+			"server_version_num", serverVersionNum,
+			"equivalent", config.PostgresExtractDatePartEquivalent,
+		)
+	}
 	d.generatorConfig = config
 	// Sync TargetSchema to d.config for backward compatibility
 	// (other methods read from d.config.TargetSchema)
 	d.config.TargetSchema = config.TargetSchema
+}
+
+func extractDatePartEquivalent(serverVersionNum int) bool {
+	return serverVersionNum < 140000
 }
 
 // refreshDefaultOperatorClasses reloads the default operator class of every access method, keyed by
@@ -530,8 +546,6 @@ func (d *PostgresDatabase) views() ([]string, error) {
 		if d.config.TargetSchema != nil && !slices.Contains(d.config.TargetSchema, schema) {
 			continue
 		}
-		// Normalize PostgreSQL-specific syntax for generic parser compatibility
-		definition = normalizeDatePartToExtract(definition)
 		ddls = append(
 			ddls, fmt.Sprintf(
 				"CREATE VIEW %s.%s AS %s;", d.quoteIdentifierIfNeeded(schema), d.quoteIdentifierIfNeeded(name), definition,
@@ -601,8 +615,6 @@ func (d *PostgresDatabase) materializedViews() ([]string, error) {
 		definition = strings.ReplaceAll(definition, "\n", " ")
 		definition = suffixSemicolon.ReplaceAllString(definition, "")
 		definition = spaces.ReplaceAllString(definition, " ")
-		// Normalize PostgreSQL-specific syntax for generic parser compatibility
-		definition = normalizeDatePartToExtract(definition)
 		ddls = append(
 			ddls, fmt.Sprintf(
 				"CREATE MATERIALIZED VIEW %s.%s AS %s;", d.quoteIdentifierIfNeeded(schema), d.quoteIdentifierIfNeeded(name), definition,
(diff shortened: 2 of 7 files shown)


schema/normalize.go
```diff
@@ -673,27 +709,38 @@ func normalizeExpr(expr parser.Expr, mode GeneratorMode) parser.Expr {
...
-			Qualifier: e.Qualifier,
-			Name:      normalizedName,
-			Distinct:  e.Distinct,
-			Exprs:     normalizedExprs,
-			Over:      e.Over,
+			Qualifier:   e.Qualifier,
+			Name:        normalizedName,
+			Distinct:    e.Distinct,
+			Exprs:       normalizedExprs,
+			WithinGroup: normalizeOrderBy(e.WithinGroup, mode, postgresExtractDatePartEquivalent),
```
moznion, line 734: Comparing WITHIN GROUP is the right call, but the implicit casts PostgreSQL renders into the definition aren't stripped, so this breaks idempotency for ordered-set aggregates:

```sql
-- what the user writes (amount is int)
SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY amount) AS median FROM events;

-- what pg_get_viewdef() returns
SELECT percentile_cont(0.5::double precision)
  WITHIN GROUP (ORDER BY (amount::double precision)) AS median FROM events;
```

master dropped WithinGroup entirely, so these happened to compare equal.  Now they don't, and every apply emits DROP VIEW + CREATE VIEW. Any view using percentile_cont / percentile_disc / mode() over an int or numeric column is affected, and DROP VIEW fails when dependents exist.

This case fails on PG13 and PG18 here, and passes on master:

```yaml
ViewWithOrderedSetAggregateOverIntColumnIsIdempotent:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE events (
      amount int
    );
    CREATE VIEW event_medians AS
      SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY amount) AS median
      FROM events;
  desired: |
    CREATE TABLE events (
      amount int
    );
    CREATE VIEW event_medians AS
      SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY amount) AS median
      FROM events;
  up: ""
  down: ""

  # Current schema is not idempotent. Expected no changes when reapplying
  # current schema, but got:
  #   DROP VIEW public.event_medians;
  #   CREATE VIEW public.event_medians AS select percentile_cont(0.5)
  #     within group( order by amount asc) as median from events;
```

The ORDER BY inside WITHIN GROUP needs the same cast stripping used elsewhere (stripFuncResultTextCasts / isFoldableValueCastTypeName). The double precision column case already passes, so casts are the only remaining gap.

schema/normalize.go
```diff
@@ -673,27 +709,38 @@ func normalizeExpr(expr parser.Expr, mode GeneratorMode) parser.Expr {
...
+			Over:        normalizeOver(e.Over, mode, postgresExtractDatePartEquivalent),
```
moznion, line 735: Normalizing Over and adding the ExtractExpr case are version-independent changes unrelated to EXTRACT/date_part. Comparing against master, both fix real bugs: 

- OVER was passed through unnormalized, so OVER (PARTITION BY (a)) and OVER (PARTITION BY a) compared as different and recreated the view.
- ExtractExpr.Source was unnormalized, so EXTRACT(year FROM (x)) and EXTRACT(year FROM x) compared as different.

Worth keeping, but please

1. mention them in the description and changelog
2. add tests on the equivalent=false path — there are none today

Both cases below fail on master and pass here, with no version gate, so they run on both the PG13 and PG18 jobs:

```yaml
ViewWindowPartitionParenthesesAreEquivalent:
  legacy_ignore_quotes: false                                                                            
  current: |
    CREATE TABLE events (
      created_at timestamp,
      amount int
    );
    CREATE VIEW event_totals AS
      SELECT sum(amount) OVER (PARTITION BY created_at) AS running_total
      FROM events;
  desired: |
    CREATE TABLE events (
      created_at timestamp,
      amount int
    );
    CREATE VIEW event_totals AS
      SELECT sum(amount) OVER (PARTITION BY (created_at)) AS running_total
      FROM events;
  up: ""
  down: ""

ViewExtractSourceParenthesesAreEquivalent:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE events (
      created_at timestamp
    );                                                                                                   
    CREATE VIEW event_years AS
      SELECT EXTRACT(year FROM created_at) AS event_year
      FROM events;
  desired: |
    CREATE TABLE events (
      created_at timestamp
    );
    CREATE VIEW event_years AS
      SELECT EXTRACT(year FROM (created_at)) AS event_year
      FROM events;
  up: ""
  down: ""
```

Note the ExtractExpr field lowercasing and Source recursion also apply to MySQL mode, and no mysqldef test was added. Is that intended?

schema/normalize.go
```diff
@@ -611,9 +640,16 @@ func stripFuncResultTextCasts(expr parser.Expr) parser.Expr {
 // normalizeExpr normalizes an expression.
 // This is similar to normalizeCheckExpr but tailored for other contexts.
 func normalizeExpr(expr parser.Expr, mode GeneratorMode) parser.Expr {
```
moznion, line 642: `normalizeExpr`, `normalizeSelectExpr` and `normalizeExprPreservingQualifiers` remain as thin wrappers hardcoding false. Forgetting to thread the flag compiles fine and silently degrades to false.

Passing a small `normalizeOptions` struct (or `GeneratorConfig`) alongside mode would make a missing flag a compile error and avoid threading a bool through 12+ signatures the next time an engine-specific flag is added.

moznion: at `normalizePrivilegesForComparison`, it seems this degrade occurs.
`shouldDropAndCreateView()` uses this to decide whether `CREATE OR REPLACE VIEW` is possible, but `normalizeSelectExpr` (false) treats date_part and `EXTRACT` as different. On PG13, adding a single column to a view that uses `EXTRACT` emits `DROP + CREATE` instead of `CREATE OR REPLACE`. PG18 emits `CREATE OR REPLACE`.

master behaves the same, so it's not a regression from this PR, but it's the exact gap this PR is closing. Passes on PG18, fails on PG13:

Repro:

```yaml
ViewColumnAdditionWithExtractUsesReplace:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE events (
      created_at timestamp,
      amount int
    );                                                                                                   
    CREATE VIEW event_years AS
      SELECT EXTRACT(year FROM created_at) AS event_year
      FROM events;
  desired: |
    CREATE TABLE events (
      created_at timestamp,
      amount int
    );
    CREATE VIEW event_years AS
      SELECT EXTRACT(year FROM created_at) AS event_year, amount AS event_amount
      FROM events;
  up: |
    CREATE OR REPLACE VIEW public.event_years AS select EXTRACT(year FROM created_at) as event_year, amount as event_amount from events;
  down: |
    DROP VIEW public.event_years;
    CREATE VIEW public.event_years AS select EXTRACT(year FROM created_at) as event_year from events;
```

database/postgres/database.go
```diff
@@ -56,12 +56,28 @@ func (d *PostgresDatabase) SetGeneratorConfig(config database.GeneratorConfig) {
 		d.defaultOpclasses = map[string]bool{}
 	}
 	config.PostgresDefaultOperatorClasses = d.defaultOpclasses
+	config.PostgresExtractDatePartEquivalent = false
+	var serverVersionNum int
+	if err := d.db.QueryRow("SHOW server_version_num").Scan(&serverVersionNum); err != nil {
+		slog.Debug("Failed to get PostgreSQL server_version_num", "error", err)
```
moznion, line 62: Debug is close to swallowing this. If the lookup fails against an actual PG13, the symptom is "every view using EXTRACT is recreated on every run", which is loud enough to warrant slog.Warn.

cmd/psqldef/tests_views.yml
```diff
@@ -493,6 +493,87 @@ ViewWithGroupByAndHaving:
...
+      created_at timestamp
+    );
+    CREATE VIEW event_years AS
+      SELECT
+        date_part('year', created_at) AS event_year,
+        pg_catalog.date_part('month', created_at) AS event_month
+      FROM events;
+  up: ""
+  down: ""
+ExtractAndDatePartEquivalentInWindowAndWithinGroupBeforePostgres14:
```
moznion, line 519: max_version: "13" means this never runs on the PG18 job, but the window / WITHIN GROUP normalization itself is version-independent. Please add a case that runs on PG14+ too.

cmd/psqldef/tests_views.yml
```diff
@@ -493,6 +493,87 @@ ViewWithGroupByAndHaving:
...
+    );
+    CREATE VIEW event_years AS
+      SELECT date_part('year', created_at) AS event_year
+      FROM events;
+  up: |
+    DROP VIEW public.event_years;
+    CREATE VIEW public.event_years AS select date_part('year', created_at) as event_year from events;
+  down: |
+    DROP VIEW public.event_years;
+    CREATE VIEW public.event_years AS select EXTRACT(year FROM created_at) as event_year from events;
```
moznion, line 576: The following test case fails on master (generated DDL is empty — the change is silently missed) and passes on this branch. It uses double precision columns so no implicit cast is involved, keeping it independent of the issue in normalization:

```yaml
ViewWithinGroupOrderByChangeIsDetected:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE events (
      amount double precision,
      weight double precision
    );
    CREATE VIEW event_medians AS
      SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY amount) AS median
      FROM events;
  desired: |
    CREATE TABLE events (
      amount double precision,
      weight double precision
    );
    CREATE VIEW event_medians AS
      SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY weight) AS median
      FROM events;
  up: |
    DROP VIEW public.event_medians; CREATE VIEW public.event_medians AS select percentile_cont(0.5) within group( order by weight asc) as
median from events;
  down: |
    DROP VIEW public.event_medians;
    CREATE VIEW public.event_medians AS select percentile_cont(0.5) within group( order by amount asc) as median from events;
```

database/postgres/database_test.go
```diff
@@ -48,6 +48,41 @@ func TestUnixSocketConnection(t *testing.T) {
...
+		CREATE MATERIALIZED VIEW event_years AS
+		SELECT date_part('year', created_at) AS event_year FROM events;
+	`)
+	require.NoError(t, err)
+
+	db.SetGeneratorConfig(database.GeneratorConfig{LegacyIgnoreQuotes: false})
+	exported, err := db.ExportDDLs()
+	require.NoError(t, err)
+	assert.Contains(t, exported, "date_part(")
+	assert.NotContains(t, exported, "EXTRACT(")
```
moznion, line 83: NotContains over the whole export will break once this test database grows another object containing EXTRACT. Assert on the materialized view's own definition instead.

schema/generator_test.go
```diff
@@ -465,14 +521,108 @@ func TestNormalizeViewDefinition(t *testing.T) {
...
+				"date_part",
+				tt.exprs,
+			)
+			assert.False(t, ok)
+			assert.Nil(t, extract)
+		})
+	}
+}
+
+func TestGeneratePostgresViewExtractDatePartComparison(t *testing.T) {
```
moznion, line 588: Good to see both equivalent: true and false covered. Please also pin the WITHIN GROUP and OVER normalization with equivalent = false — nothing currently guards that side of the three behavior changes this PR makes.

database/postgres/database.go
```diff
@@ -56,12 +56,28 @@ func (d *PostgresDatabase) SetGeneratorConfig(config database.GeneratorConfig) {
...
+		)
+	}
 	d.generatorConfig = config
 	// Sync TargetSchema to d.config for backward compatibility
 	// (other methods read from d.config.TargetSchema)
 	d.config.TargetSchema = config.TargetSchema
 }
 
+func extractDatePartEquivalent(serverVersionNum int) bool {
+	return serverVersionNum < 140000
```
moznion, line 78: PostgreSQL-compatible servers land here too. Redshift reports an 8.x-era server_version_num, so it gets equivalent = true automatically. Is that intended?

Also: 140000 deserves a named constant, and extractDatePartEquivalent doesn't say what is equivalent to what — maybe datePartIsExtractAlias?

moznion (commented):
Thank you for your contribution! I left some comments.

PR #1349, opened 2026-09-02
Fix TRIM expression normalization for CHECK constraints

wreulicke:
## Summary

Fix normalization and parsing of `TRIM` expressions used in `CHECK` constraints.

## Problem

PostgreSQL may export an expression such as:

```sql
char_length(TRIM(BOTH FROM value)) > 0
```

while the desired schema contains the equivalent:

```sql
char_length(trim(value)) > 0
```

These expressions were treated as different during repeated schema application, causing unnecessary `CHECK` constraint changes.

MySQL keyword forms such as `TRIM(BOTH ' ' FROM value)` were also not fully supported by the parser.

## Changes

- Normalize `trim` and PostgreSQL `btrim` function calls to `TrimExpr`.
- Support `TRIM(BOTH|LEADING|TRAILING [trim_character] FROM expression)`.
- Treat explicit `BOTH` as the default trim direction.
- Preserve explicit `LEADING` and `TRAILING` directions.
- Add parser, normalization, and MySQL/PostgreSQL regression tests.

## Tests

- `go test ./parser ./schema`
- `go test ./cmd/mysqldef -run '^TestApply$/TrimCheckConstraintIsIdempotent$' -count=1`
- `go test ./cmd/psqldef -run '^TestApply$/UnnamedTrimCheckConstraintIsIdempotent$' -count=1`


moznion (changes requested):
Thank you for your contribution!

I think there’s an issue with how trim is handled in the generated DDL.

For example, the following test case should pass:

```yaml
TwoArgumentBtrimKeepsArgumentOrder:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE test (
      t1 varchar(255) NOT NULL
    );
  desired: |
    CREATE TABLE test (
      t1 varchar(255) NOT NULL,
      CONSTRAINT test_t1_chk CHECK (btrim(t1, 'xy') <> '')
    );
  up: |
    ALTER TABLE public.test ADD CONSTRAINT test_t1_chk CHECK (trim('xy' from t1) <> '');
  down: |
    ALTER TABLE public.test DROP CONSTRAINT test_t1_chk;
```

However, it currently generates `ALTER TABLE public.test ADD CONSTRAINT test_t1_chk CHECK (trim(t1 from 'xy') <> '')`, in this output, the arguments to trim are reversed.

I think this pull request should also include test cases covering the following:

- Generated DDL for `btrim`, `ltrim`, and `rtrim`, with both the `up` and `down` directions explicitly specified.
  - Both the one-argument and two-argument forms of each function.

wreulicke:
@moznion 
Thanks for catching this!
I've changed the following things. Could you check the changes again? 

- Fixed the swapped TrimChar/String in the two-argument btrim/ltrim/rtrim normalization.
- Fixed the existing unit test that had the same argument order wrong (so it didn't catch the bug).
- Added a test case covering one- and two-argument btrim/ltrim/rtrim with explicit up/down,

cmd/psqldef/tests_constraints.yml
```diff
@@ -350,6 +350,56 @@ CheckConstraint:
     ALTER TABLE "public"."test" DROP CONSTRAINT "test_t4_check";
     ALTER TABLE "public"."test" DROP CONSTRAINT "test_d1_check";
     ALTER TABLE "public"."test" DROP CONSTRAINT "test_d2_check";
+UnnamedTrimCheckConstraintIsIdempotent:
```
moznion: [question] Is this test case necessary? It seems to only verify that the syntax is correct, and `TrimFunctionCheckConstraintsKeepArgumentOrder` might already cover that. If we keep this test, it would be nice to also verify the up/down migration.

cmd/mysqldef/tests_constraints.yml
```diff
@@ -11,6 +11,17 @@ ConstraintCheck:
...
+  desired: |
+    CREATE TABLE `test` (
+      `t1` varchar(255) NOT NULL,
+      `t2` varchar(255) NOT NULL,
+      `t3` varchar(255) NOT NULL,
+      CHECK (CHAR_LENGTH(TRIM(BOTH ' ' FROM `t1`)) > 0),
+      CHECK (CHAR_LENGTH(TRIM(LEADING ' ' FROM `t2`)) > 0),
+      CHECK (CHAR_LENGTH(TRIM(TRAILING ' ' FROM `t3`)) > 0)
+    ) ENGINE=InnoDB;
+  min_version: '8.0'
```
moznion: I think mysqldef should also have test cases like the ones in postgres. What do you think?

moznion (commented):
I left two comments.

wreulicke:
@moznion 
Thanks! I've fixed it:

- Removed UnnamedTrimCheckConstraintIsIdempotent — unnamed check-constraint matching is already covered elsewhere.
- TrimFunctionCheckConstraints now covers its trim keyword form too, with explicit up/down for all three engines.
  - btrim is Postgres-specific, so it's only tested there.

moznion (approved):
I appreciate your contribution!

PR #1351, opened 2026-09-07
psqldef: allow LANGUAGE as an unquoted identifier in the generic parser

e-koma:
Fixes #1345

## Summary

PostgreSQL treats `LANGUAGE` as non-reserved, so `--export` writes index columns as `USING btree (tenant_id, language)` without quotes. The generic parser treated `language` as the `LANGUAGE` token, but did not classify it as an identifier, so it could not parse the schema it exported.

This adds `LANGUAGE` to `non_reserved_keyword`, same approach as `INDEX` / `TRIGGER` (#1282). `CREATE FUNCTION ... LANGUAGE sql` is unchanged.

On v3, that parse failure also forced a file-wide pgquery fallback, which dropped quotes from reserved identifiers such as `"group"`. Fixing the generic parser removes that fallback too.

## Test plan

- [x] `make parser`
- [x] `PSQLDEF_PARSER=generic go test ./parser`
- [x] `PSQLDEF_PARSER=generic go test ./cmd/psqldef`


moznion (approved):
@e-koma Thank you for your contribution. This seems good. Could you please regenerate the `parser/parser.go` to resolve a merge conflict?

e-koma:
@moznion 
Thanks !! I've merged upstream `master` and regenerated `parser/parser.go` !!

moznion:
I appreciate your work.

PR #1352, opened 2026-09-07
sqlite3def: accept and ignore DELETE / PRAGMA statements from Cloudflare D1 schema exports

knaka:
Cloudflare D1's DDL is almost equivalent to SQLite3's one but its schema export (`wrangler d1 export`, even with `--no-data` option) interleaves non-DDL, data initialization statements with the schema DDL:

```sql
PRAGMA defer_foreign_keys=TRUE;

CREATE TABLE ...;

...

DELETE FROM sqlite_sequence;
```

# Changes

- `DELETE`: `delete_statement` already existed in the grammar but was only
  reachable inside trigger bodies. Wire it into the top-level `statement` rule.
- `PRAGMA`: add a `PRAGMA` keyword token and a small `pragma_statement`
  rule (`PRAGMA name`, `PRAGMA name = value`, `PRAGMA name(value)`) plus a
  `Pragma` AST node. The value is not interpreted.
- `database.GenericParser.Parse`: skip parsed `*parser.Delete` and
  `*parser.Pragma` statements so they never reach the schema differ.
- Add tests.


k0kubun (approved):
seems fine to support that syntax.

I nominated you as a maintainer who's primarily in charge of sqlite3def, so you're free to change what's supposed in sqlite3def as you wish :)

knaka:
@k0kubun Thanks for the review, and for the trust. This one touched the generic parser, so I wanted a second pair of eyes given the wider blast radius. Going forward, for changes that are specific to sqlite3def, I'll go ahead and merge them myself.

PR #1353, opened 2026-09-08
mysqldef: support TDSQL schema extensions

Dirkkroos:
Add TDSQL compatibility on top of the existing MySQL/MariaDB/TiDB support in mysqldef, covering the schema-level extensions TDSQL adds over vanilla MySQL: table/database distribution and sync-level policies, TTL with archive tables, storage tiers, global secondary indexes (GSI/ICI) with independent partitioning, and TDSQL's own partition-management DDL.

- parser: add TDSQL keywords (DISTRIBUTION, SYNC_LEVEL, TTL, ALGORITHM on GLOBAL indexes, STORAGE_TIER, PARTITION POLICY, DISTRIBUTION POLICY, VISIBLE/INVISIBLE, etc.) as non-reserved where TDSQL itself treats them as such, following the existing MSSQL/TiDB precedent of extending the shared grammar rather than forking a dialect-specific parser
- schema/ast, schema/parser: represent GSI/ICI, independent GSI partitioning, table/database DISTRIBUTION and SYNC_LEVEL, TTL options, and PARTITION POLICY / DISTRIBUTION POLICY objects as first-class fields alongside the existing MSSQL-specific fields on Index and TablePartition
- schema/generator: diff and idempotent DDL generation for all of the above, including:
  - GSI ALGORITHM handling: an index-level or config-level ALGORITHM is preserved verbatim; when neither is specified sqldef no longer forces ALGORITHM = COPY, leaving the choice to TDSQL
  - RANGE/LIST partition add/drop/reorganize and INTERVAL partitioning diffed against TDSQL's normalized export format
  - PARTITION POLICY / DISTRIBUTION POLICY create/alter/rename/drop, diffed against the policy DDL TDSQL exposes via information_schema, independent of the tables that reference them
  - TTL, STORAGE_TIER and DISTRIBUTION/SYNC_LEVEL table options, comparing effective values rather than raw strings so options TDSQL defaults or normalizes don't produce spurious DDL
- database/mysql: export CREATE/ALTER DATABASE alongside table DDL when diffing against a live TDSQL instance (ExportDDLsForDiff), while leaving `--export` scoped to the selected database as before; export and bind PARTITION POLICY / DISTRIBUTION POLICY definitions
- cmd-mysqldef.md: document the TDSQL Compatibility surface (schema extensions, policies, GSI/ICI, known ALTER-is-silently-ignored caveats for DISTRIBUTION/SYNC_LEVEL)
- cmd/mysqldef/tests_tdsql.yml: 271 offline test cases (parser, diff, idempotency) covering the above; verified additionally with a full online regression against a real TDSQL instance (271/271 passed) and a full MySQL/MariaDB/TiDB regression (313/313 passed) to confirm no regressions in existing dialects


moznion (commented):
@Dirkkroos Thank you for your contribution. Some CI checks (especially the PostgreSQL ones) are failing. Could you please address them?

PR #1354, opened 2026-09-11
parser: disable goyacc line directives

gfx:
Source line changes in `parser/parser.y` update many `//line` directives in the generated parser, creating unnecessary diff churn and merge conflicts. Pass `-l` to goyacc and regenerate `parser/parser.go` to omit these directives.

Verified that generated code is unchanged apart from line directives, the generation header, and blank lines. Grammar conflict counts are unchanged (28 shift/reduce, 545 reduce/reduce).

Validation:
- `make build`, `make format`, and `make lint` passed.
- `PSQLDEF_PARSER=generic go test ./parser ./cmd/sqlite3def` passed on an isolated rerun; an initial overlapping SQLite test run hit database locking.
- `PSQLDEF_PARSER=generic make test-all-flavors` was attempted but stopped due to unavailable MySQL/SQL Server services and PostgreSQL authentication failures; flavor-specific runs were not reached.


moznion (approved):
It seems reasonable. LGTM.

PR #1356, opened 2026-09-13
parser: drop the never-reduced duplicate rule in function_call_generic

178inaba:
## Why

`goyacc` reports `function_call_generic: sql_id '(' select_expression_list ')'` as `never reduced`. The same input is also accepted by `sql_id '(' select_expression_list ')' over_expression`, because `over_expression` can be empty. yacc resolves the reduce/reduce conflict in favor of the earlier rule, which is the empty `over_expression` one, so the plain rule is unreachable for every input.

The ambiguity still costs one reduce/reduce conflict per token in the follow set, 201 in total, and `AGENTS.md` says no reduce/reduce conflicts are allowed.

## What

Drop the unreachable rule. Both alternatives build `&FuncExpr{Name: $1, Exprs: $3}` (with a nil `Over` on the remaining one), and the removed one was never reduced, so the accepted SQL, the resulting AST, and syntax error messages and positions are all unchanged. `make parser` no longer reports a rule as `never reduced`, reduce/reduce drops from 487 to 286, and shift/reduce stays at 28.

`TestFunctionCallForms` pins that each function call form still parses and renders back to the same SQL: with and without arguments, `DISTINCT`, `OVER (...)`, `WITHIN GROUP (...)`, `LAG` / `LEAD` with and without `OVER`, and schema-qualified calls such as `s.f(x)`. There were no parser tests covering these forms before.

<details>
<summary>Verification</summary>

In addition to the test suites, 148 statements (function call forms, operators from the follow set, `DEFAULT` / `CHECK` / index expression contexts, and malformed calls) were run through the parser in all four modes (mysql, postgres, sqlite3, mssql) before and after the change, recording either `String()` output or the full error message. The two dumps are byte-identical.

</details>


moznion (approved):
LGTM

PR #1358, opened 2026-09-15
psqldef: treat MySQL-only UNUSED keywords as identifiers

dim0627:
## Summary

The generic parser rejects a column named `year_month` in PostgreSQL mode:

```sql
CREATE TABLE public.t (id bigint NOT NULL, year_month text NOT NULL);
CREATE INDEX t_ym ON public.t USING btree (id, year_month);
```

```
WARN Generic parser failed on full SQL, using pgquery fallback error="... syntax error at line 1, column 55 near 'year_month'"
```

PostgreSQL does not reserve `year_month`, so `pg_get_indexdef` and `psqldef --export` write it unquoted. Today pgquery rescues the file, but this is the kind of fallback `AGENTS.md` treats as a bug.

## Cause

`parser/token.go` maps MySQL-only keywords to `UNUSED`. Examples are INTERVAL units (`year_month`, `day_hour`, ...), `xor`, and `distinctrow`. In the grammar, `UNUSED` is accepted only by `table_id` (#1125) and `privilege` (#1285). So these words work as table names, but not as column names, index columns, or column references.

## Fix

Lex `UNUSED` words as plain `ID` tokens in PostgreSQL mode. This is the same lexer-split approach as `PG_KEY` (#1202) and `PG_COMMENT` (#1291). The exceptions are `BOTH`, `LEADING` and `TRAILING`, which PostgreSQL reserves. They are the only `UNUSED` words in PostgreSQL's reserved list.

The grammar is not changed, and `make parser` output is identical. MySQL, SQLite3 and SQL Server modes behave exactly as before.

I first tried adding `UNUSED` to `non_reserved_keyword`, like #1351 did for `LANGUAGE`. It kept the conflict counts, but the grammar is shared, so it also changed MySQL mode. For example, `SELECT DISTINCTROW a FROM t` parsed as `select distinctrow as a from t`, and `INTERVAL '1-2' YEAR_MONTH` parsed as an alias. Before, both were syntax errors. It also let `year_month` and `both` through as unquoted column names where the database rejects them. Splitting in the lexer avoids all of this.

One side effect in PostgreSQL mode: `int4`, `int8`, `float4` and `float8` are also `UNUSED`, so the generic parser now reads them as type names. Before, they failed and fell back to pgquery. Existing type normalization maps them to `integer` and so on. Against PostgreSQL 18, a table with an `integer` column and a desired `int4` column gives `-- Nothing is modified --`.

## Tests

- `parser/parser_test.go`: `TestUnusedKeywordsAsUnquotedIdentifiers`
  - Cases that must parse in PostgreSQL mode: `year_month` / `xor` in column definitions, index columns, a partial index predicate, and a UNIQUE constraint.
  - Cases that must still fail: `year_month` as a column in MySQL mode, and `both` as a column in PostgreSQL mode.
- `cmd/psqldef/tests_indices.yml`: `CreateIndexOnColumnNamedYearMonth` adds and drops an index on a `year_month` column against a live PostgreSQL.

Both tests fail on `master` and pass here.

## Verification

- [x] `make parser` (no diff; 28 shift/reduce, 487 reduce/reduce)
- [x] `make build`, `make lint`, `make format`
- [x] `go test ./...` with PostgreSQL 18 (`PSQLDEF_PARSER=generic`), MySQL 8.0 and SQL Server 2025
- [ ] MariaDB, TiDB and pgvector flavors were not run locally. The change only affects PostgreSQL-mode lexing.


moznion (approved):
LGTM, thank you for your contribution.

PR #1360, opened 2026-09-15
psqldef: parse ALTER TABLE ONLY UNIQUE constraints in the generic parser

e-koma:
## Summary

The generic parser can parse `ALTER TABLE ONLY ... ADD CONSTRAINT` for `FOREIGN KEY` and `EXCLUDE`, but it cannot parse it for `UNIQUE`. PostgreSQL and `pg_dump` write unique constraints in this form. psqldef reads these schemas, so it should be able to parse them.

This change adds the `ONLY` variant to the existing `UNIQUE` rule. `ALTER TABLE ... UNIQUE` without `ONLY` does not change.

## Environment

- psqldef v3.11.21 (`PSQLDEF_PARSER=generic`)

## How to reproduce

`repro.sql`:

```sql
CREATE TABLE public.dummy (
    id bigint NOT NULL,
    column_a bigint NOT NULL,
    column_b bigint NOT NULL
);

ALTER TABLE ONLY public.dummy ADD CONSTRAINT dummy_uniq UNIQUE (column_a, column_b);
```
```
$ echo "-- empty" > empty.sql
$ PSQLDEF_PARSER=generic psqldef empty.sql < repro.sql
found syntax error when parsing DDL "ALTER TABLE ONLY public.dummy ADD CONSTRAINT dummy_uniq UNIQUE (column_a, column_b)": syntax error at line 1, column 64 near 'unique'
  ALTER TABLE ONLY public.dummy ADD CONSTRAINT dummy_uniq UNIQUE (column_a, column_b)
                                                                 ^
```

The same statement without `ONLY` already parses:
```sql
ALTER TABLE public.dummy ADD CONSTRAINT dummy_uniq UNIQUE (column_a, column_b);
```

## Expected behavior
`ALTER TABLE ONLY ... ADD CONSTRAINT ... UNIQUE (...)` should parse.
PostgreSQL accepts it, and `pg_dump` writes it this way.

## Changes

- `parser/parser.y`: add an `ALTER TABLE ONLY ... ADD CONSTRAINT ... UNIQUE` rule next to the existing UNIQUE rule

## Test plan

- [x] `make parser`
- [x] `PSQLDEF_PARSER=generic go test ./parser`
- [x] `PSQLDEF_PARSER=generic go test ./cmd/psqldef`


moznion (approved):
Thank you for your contribution. Looks good.

PR #1361, opened 2026-09-15
mysqldef: preserve parentheses around literal DEFAULT values

zaneli:
fix #1357

## Problem

For MySQL `TEXT`/`BLOB`/`JSON`/`GEOMETRY` columns, MySQL rejects a bare literal `DEFAULT` (error 1101) and only accepts it via the explicit parenthesized "expression default" form — even for a plain literal like a string:

```sql
-- Rejected by MySQL:
note TEXT NOT NULL DEFAULT 'foo'
```

```sql
-- Required instead:
note TEXT NOT NULL DEFAULT ('foo')
```

Previously, mysqldef silently flattened DEFAULT ('foo') in the desired schema down to DEFAULT 'foo' when generating DDL, which produced SQL that MySQL rejects on --apply.

Root cause

The shared generic parser's default_value_expression rule for '(' default_value_expression ')' discarded the parentheses at parse time ($$ = $2), so ('foo') and 'foo' produced an identical AST node with no way for the generator (schema/generator.go) to tell them apart.

Fix

- parser/parser.y: wrap the inner expression in &ParenExpr{...} instead of discarding the parentheses, so the AST preserves whether the user wrote a literal default with or without parens.
- schema/generator.go:
  - generateDefaultDefinition re-emits DEFAULT(literal) only when g.mode == GeneratorModeMysql; every other dialect keeps unwrapping the parentheses to preserve its existing (byte-identical) output.
  - isNullDefault and areSameDefaultValue (via new unwrapParenExpr / unwrapCastAndParens helpers) are made paren-tolerant, so idempotency comparisons (current vs desired) are unaffected by this wrapping change.

Testing

- Added TextColumnWithParenthesizedStringDefault and AddColumnWithParenthesizedStringDefault to cmd/mysqldef/tests_datatypes.yml.
- Verified against MySQL and PostgreSQL, plus the sqlite3def/parser/schema unit test suites — no regressions found.
- mssqldef wasn't verified against a live SQL Server instance, but its own database/mssql/parser.go (normalizeDefaultExpressions) already strips this kind of parentheses at the text level before the shared parser sees it, so it should be unaffected.


moznion (commented):
Thanks for the fix! The original issue is resolved but seemingly this PR introduces another issue.

On master these defaults were emitted without parentheses, so MySQL stored them as plain literal defaults. With this PR they are emitted as `DEFAULT(current_timestamp)` / `DEFAULT(true)`. MySQL stores those as expression defaults and exports them as `(now())` / `(true)`, which the current comparison logic doesn't treat as equal to the desired schema. As a result, every run generates a `CHANGE COLUMN`.


Minimal Repro (idempotency-only, `cmd/mysqldef/tests_datatypes.yml`):

```yaml
AddColumnWithParenthesizedCurrentTimestampDefault:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE hello_default (
      id int NOT NULL
    );
  desired: |
    CREATE TABLE hello_default (
      id int NOT NULL,
      created_at datetime DEFAULT (CURRENT_TIMESTAMP)
    );
  min_version: '8.0.13'
AddColumnWithParenthesizedBooleanDefault:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE hello_default (
      id int NOT NULL
    );
  desired: |
    CREATE TABLE hello_default (
      id int NOT NULL,
      enabled tinyint(1) DEFAULT (true)
    );
  min_version: '8.0.13'
```

Keep the parentheses only for the literal kinds that need them, e.g. strings (and possibly numbers). Keep emitting `ValArg` (e.g., `CURRENT_TIMESTAMP`, `NULL`) and bool defaults without parentheses, as before. It would also be good to add the tests above, plus a CHANGE COLUMN case like `DEFAULT ('foo')` -> `DEFAULT ('bar')` on a TEXT column.

The underlying comparison gap (`(now())` vs `CURRENT_TIMESTAMP`, `(true)` vs `1`) already exists on master when such columns are created via `CREATE TABLE`, so I think it can be handled in a separate issue/PR.

moznion (commented):
Thanks for the fix! The original issue is resolved but seemingly this PR introduces another issue.

On master these defaults were emitted without parentheses, so MySQL stored them as plain literal defaults. With this PR they are emitted as `DEFAULT(current_timestamp)` / `DEFAULT(true)`. MySQL stores those as expression defaults and exports them as `(now())` / `(true)`, which the current comparison logic doesn't treat as equal to the desired schema. As a result, every run generates a `CHANGE COLUMN`.


Minimal Repro (idempotency-only, `cmd/mysqldef/tests_datatypes.yml`):

```yaml
AddColumnWithParenthesizedCurrentTimestampDefault:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE hello_default (
      id int NOT NULL
    );
  desired: |
    CREATE TABLE hello_default (
      id int NOT NULL,
      created_at datetime DEFAULT (CURRENT_TIMESTAMP)
    );
  min_version: '8.0.13'
AddColumnWithParenthesizedBooleanDefault:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE hello_default (
      id int NOT NULL
    );
  desired: |
    CREATE TABLE hello_default (
      id int NOT NULL,
      enabled tinyint(1) DEFAULT (true)
    );
  min_version: '8.0.13'
```

Keep the parentheses only for the literal kinds that need them, e.g. strings (and possibly numbers). Keep emitting `ValArg` (e.g., `CURRENT_TIMESTAMP`, `NULL`) and bool defaults without parentheses, as before. It would also be good to add the tests above, plus a CHANGE COLUMN case like `DEFAULT ('foo')` -> `DEFAULT ('bar')` on a TEXT column.

The underlying comparison gap (`(now())` vs `CURRENT_TIMESTAMP`, `(true)` vs `1`) already exists on master when such columns are created via `CREATE TABLE`, so I think it can be handled in a separate issue/PR.

zaneli:
@moznion 
Thanks for the review.

I've addressed the feedback. In addition to the CHANGE COLUMN test for the TEXT type, I also added an idempotency test for an int column with a parenthesized default value. 23e9668

moznion:
I appreciate your contribution!

PR #1362, opened 2026-09-16
psqldef: diff a renamed table's privileges under its new name

moznion:
skipping the orphan check for it. That hid a second problem: the carried-over privileges stay in currentPrivileges under the old name, so nothing diffs them against the desired GRANTs, which are written under the new name. A privilege removed in the same run as the rename was therefore never revoked, and the apply left drift behind:

```sql
-- current: GRANT SELECT, INSERT, UPDATE ON old_logs TO app_user
-- desired: new_logs (@renamed from=old_logs), GRANT SELECT, INSERT
ALTER TABLE "public"."old_logs" RENAME TO "new_logs";
-- UPDATE is still granted; the next run emits REVOKE UPDATE.
```

PostgreSQL keeps the privileges on the renamed table, so they are the current state of the new name. Re-point them when the rename is generated instead of excluding them from the diff: the normal privilege diff then revokes what the desired schema dropped, grants only what it added, and emits nothing when the grants are unchanged. The orphan check now revokes under the new name too, so the rename skip is gone and only the dropped-table skip remains.

A column-level privilege needs the same treatment one level down: it stores the column name in the privilege itself ("SELECT (secret)"), and PostgreSQL rewrites that name when the column is renamed. Leaving the old name there made the diff revoke a column that no longer exists, which failed the apply outright:

```sql
ALTER TABLE "public"."logs" RENAME COLUMN "secret" TO "secret_v2";
REVOKE SELECT (secret) ON TABLE "public"."logs" FROM "app_user";
-- 42703: column "secret" of relation "logs" does not exist
```

That one predates #1337 for a plain column rename; renaming the table used to hide it for its columns by skipping the privilege check altogether. Rename the column inside the current privileges as well, when the RENAME COLUMN is generated.

Cover the rename cases for both managed_roles and manage.privilege, including the one only the latter has: a rule that forbids REVOKE reports the privilege the rename carried over as skipped, so the drift stays visible in the plan.


moznion:
cc @draftcode

PR #1364, opened 2026-09-16
parser: allow STRICT, WITHOUT and PRAGMA as identifiers

mattn:
STRICT and WITHOUT were removed from non_reserved_keyword in c2106f5 to avoid a shift/reduce conflict with SQLite table options, and PRAGMA was added as a reserved token in #1352. As a result, columns named strict, without or pragma fail to parse in every mode, even though all supported databases accept them as identifiers. This restores them as non-reserved keywords and resolves the conflict with %prec instead, so that STRICT/WITHOUT right after the column list is always shifted as a table option.


moznion (approved):
Thank you for your contribution. Unfortunately, the `parser.go` gets a merge conflict so could you please regenerate that?

mattn:
Thanks, rebased onto master and regenerated parser.go.

moznion:
I appreciate your work!

PR #1367, opened 2026-09-16
feat(sqlite3def): support bracket-quoted identifiers

knaka:
Implements #1208

Allow `[foo]` as an identifier quote in the generic parser's SQLite3 mode, matching SQLite's own compatibility handling of SQL Server-style bracket quoting (in addition to the already-supported double quotes and backticks).


moznion (approved):
LGTM, thank you.

PR #1369, opened 2026-09-16
psqldef: keep IN in generated CHECK DDL and converge OR chains on ENUM columns

178inaba:
## Why

psqldef rewrote `col IN (a, b)` into `col = ANY (ARRAY[a, b])` both when comparing CHECK constraints and when generating DDL. The two spellings are not equivalent for an ENUM column. PostgreSQL resolves the literals in an `IN` list against the type of the left operand. An `ARRAY[...]` of untyped literals is resolved on its own, to `text[]`. That causes two defects, both in CHECK constraints on ENUM columns:

1. **`IN (...)` produced DDL that PostgreSQL rejects.** Adding the constraint to an existing table failed, and the transaction rolled back, taking any unrelated DDL in the same run with it:

   ```console
   $ psqldef ... --file desired.sql repro
   -- Apply --
   BEGIN;
   ALTER TABLE "public"."orders" ADD CONSTRAINT "orders_processed_at_shape" CHECK (state = ANY (ARRAY['paid', 'shipped']) or processed_at is null);
   ROLLBACK;
   pq: operator does not exist: order_state = text (42883)
   ```

   `CREATE TABLE` emits the author's text verbatim, so this only appeared once the table already existed.

2. **An OR chain of equalities never converged.** `state = 'paid' OR state = 'shipped'` was folded into `IN (...)`, but the result skipped the `IN` → `= ANY (ARRAY[...])` conversion that a written `IN` gets. PostgreSQL stores `= ANY (ARRAY[...])`, so every run emitted `DROP CONSTRAINT` + `ADD CONSTRAINT`. `ADD CONSTRAINT ... CHECK` revalidates every row, which takes a real lock on a large table each time.

   PostgreSQL also stores an `IN` list that contains column references as an OR chain (`"Status" IN ("Fallback", 'pending')` is stored as `"Status" = "Fallback" OR "Status" = 'pending'::text`), so those constraints went down the same path.

This is the same class of bug as #1295 and #1182, for the type of the array elements rather than the shape of the operator.

## What

- `normalizeComparisonExpr` converts `IN` / `NOT IN` to `= ANY` / `<> ALL` only when comparing. Generated DDL keeps `IN (...)`, which PostgreSQL accepts for every column type and stores as the same `ScalarArrayOpExpr`. The flag that controls this is renamed from `canonicalizeArrays` to `forComparison`, because it now does more than sort arrays.
- `formatExprQuoteAware` gains a `parser.ValTuple` case. Without it, an `IN` list in quote-aware mode would fall back to `parser.String` and drop the quotes around `"Fallback"`, silently referencing a different column.
- When comparing, an OR chain folded into `IN` now goes through `normalizeComparisonExpr`, like a written `IN`.
- The OR-to-`IN` fold exists to compare the spellings PostgreSQL and SQL Server may store, so it now applies only when comparing. Generated DDL keeps the OR chain as written, consistent with keeping `IN` as written.
- `tryConvertOrChainToIn` needs two changes to support the comparison path:
  - Nested folds now arrive as `= ANY (ARRAY[...])`, so it unpacks them.
  - It no longer treats an `ANY` / `ALL` comparison as a scalar `=`. Before, it took `x = ANY (ARRAY[...])` as a single value that was the whole array.
- `tryConvertOrChainToIn` no longer sorts, since `normalizeComparisonExpr` already sorts and deduplicates the folded list.

### Changes in generated DDL

- **PostgreSQL CHECK constraints written with `IN` / `NOT IN`** are now emitted as `in (...)` / `not in (...)` instead of `= ANY (ARRAY[...])` / `<> ALL (ARRAY[...])`. This applies to `ALTER TABLE ... ADD CONSTRAINT` and to column-level CHECKs emitted with `ADD COLUMN`. The existing test expectations are updated accordingly.
- **OR chains of equalities on the same column** are emitted as written instead of being folded into a sorted `in (...)`. Before this change, a chain that contained an `= ANY (ARRAY[...])` term was folded into invalid DDL such as `x in ('c', ARRAY['a', 'b'])`.
- **Explicitly written `ANY` / `ALL`** are emitted unchanged.

None of these change what the constraint accepts, and an existing constraint does not produce a new diff.

## Not covered

- **`PSQLDEF_PARSER=pgquery`.** `database/postgres/parser.go` converts `IN` into `= ANY (ARRAY[...])` while parsing, so the fix does not apply to that path. It also runs when the generic parser fails on any statement and the whole file falls back to pgquery. pgquery is unmaintained, and that fallback is itself a bug to fix in the generic parser.
- **Partial index predicates** (`normalizeExpr`). They were not measured for the same defects.
- **`ALTER TYPE ... ADD VALUE` followed by dependent DDL in the same transaction** (`55P04`). That is a separate ordering issue.

## Tests

New YAML cases in `cmd/psqldef/tests_constraints.yml`. All of them fail before this change, with both the generic and the default parser:

- `ConstraintCheckInOnEnumColumnAdd` / `ConstraintCheckNotInOnEnumColumnAdd`: the emitted DDL failed with `42883`.
- `ConstraintCheckOrChainOnEnumColumnAdd`: the first apply succeeded, but the second run was not a no-op.
- `ConstraintCheckOrChainToInOnEnumColumn` / `ConstraintCheckInToOrChainOnEnumColumn`: a constraint stored in one spelling did not match the same predicate written in the other.
- `ConstraintCheckInColumnReferenceQuoteAware`: an `IN` list with a quoted column reference. It checks that the quotes are kept and that the constraint converges.

`schema/normalize_any_array_test.go` covers the output spelling, including an OR chain with an `ANY` term that is emitted as written.

`make test-all-flavors`, `make format` and `make lint` pass. The MSSQL suite covers the OR-chain path that `tryConvertOrChainToIn` exists for.


moznion (approved):
LGTM ✨

PR #1371, opened 2026-09-17
psqldef: accept stream as an identifier by removing the dead STREAM token

dim0627:
## Summary

The generic parser rejects `stream` as an unquoted table name in PostgreSQL mode:

```sql
CREATE INDEX idx_stream_start_at ON public.stream USING btree (id, start_at) WHERE (start_at IS NOT NULL);
```

```
WARN Generic parser failed on full SQL, using pgquery fallback error="found syntax error when parsing DDL \"CREATE INDEX idx_stream_start_at ON public.stream USING btree (id, start_at) WHERE (start_at IS NOT NULL)\": syntax error at line 1, column 51 near 'stream'"
```

PostgreSQL does not treat `stream` as a keyword, so `pg_get_indexdef` and `psqldef --export` write it unquoted. pgquery rescues the file today, but `AGENTS.md` treats this fallback as a bug.

## Cause

`parser/token.go` maps `stream` to the `STREAM` token. The only rule that used `STREAM` was the Vitess `STREAM ... FROM` statement, which e527e0c removed. Since then no grammar rule accepts the token. So `stream` fails in every position (table name, column name, index column, FK reference) and in every parser mode.

## Fix

Remove the `STREAM` token from `parser.y` and the keyword map. `stream` is now lexed as a plain `ID`.

This changes all modes, not only PostgreSQL. I think this is right, because no supported database reserves `stream`:

- PostgreSQL 18: not in `pg_get_keywords()`.
- MySQL 8.0: `information_schema.KEYWORDS` lists `STREAM` with `RESERVED = 0`.
- SQL Server 2025 and SQLite 3.51: `CREATE TABLE stream (id int PRIMARY KEY, stream int)`, `CREATE INDEX` on it, and `SELECT stream FROM stream` all succeed.

I checked the parser in all four modes before and after. For each statement below, I also parsed the same SQL with `stream` replaced by `foo`. Before the fix, every statement failed with `stream` and parsed with `foo`. After the fix, `stream` behaves exactly like `foo`:

- `CREATE TABLE stream (...)` and `CREATE TABLE s.stream (...)`
- `CREATE TABLE t (id int, stream int)`
- `CREATE INDEX i ON s.stream USING btree (id)`
- `CREATE INDEX i ON t (stream)` and a partial index with `WHERE (stream IS NOT NULL)`
- `ALTER TABLE t ADD CONSTRAINT fk FOREIGN KEY (sid) REFERENCES stream (id)`
- `CREATE VIEW v AS SELECT stream FROM stream`

I did not use the lexer split from #1358 (lex as `ID` only in PostgreSQL mode). The token is dead, so a mode check would keep a token nothing uses and leave the same bug in the other modes. I did not add `STREAM` to `non_reserved_keyword` like #1351 either, for the same reason, and because it would touch the shared grammar.

`make parser` conflict counts are unchanged: 28 shift/reduce, 286 reduce/reduce.

## Tests

- `parser/parser_test.go`: `TestStreamAsUnquotedIdentifier` parses `stream` as a table name, a column name, `CREATE INDEX ... ON public.stream`, a partial index, an index column, and an FK reference, in all four parser modes.
- `cmd/psqldef/tests_indices.yml`: `CreateIndexOnTableNamedStream` adds and drops a partial index on a `stream` table that another table references by FK, against a live PostgreSQL.

Both tests fail on `master` and pass here.

## Verification

- [x] `make parser` (conflict counts unchanged)
- [x] `make build`, `make lint`, `make format`
- [x] `go test ./...` with PostgreSQL 18 (`PSQLDEF_PARSER=generic`), MySQL 8.0 and SQL Server 2025
- [x] `psqldef --dry-run` on a real schema with a `stream` table: v3.11.22 logs the WARN above, this branch does not, and both print `-- Nothing is modified --`. Same result with `PSQLDEF_PARSER=generic`.
- [ ] MariaDB, TiDB and pgvector flavors were not run locally.


moznion (commented):
Thank you for your contribution. The overall direction of the changes looks good, but this pull request has merge conflicts. Could you please resolve them all?

moznion:
Thank you for your contribution!

PR #1373, opened 2026-09-17
parser: accept 15 SQL Server keywords as identifiers

178inaba:
The following words are lexed as keywords only for SQL Server syntax, but none of them is a reserved word in any supported database (checked against the reserved word lists of SQL Server, MySQL 8.4, PostgreSQL and SQLite):

| Syntax | Words |
|---|---|
| Built-in functions | `newid`, `newsequentialid`, `getutcdate`, `sysutcdatetime`, `try_cast` |
| Table-valued functions | `openjson`, `string_split` |
| `CROSS APPLY` / `OUTER APPLY` | `apply` |
| `COLUMNSTORE` indexes | `columnstore` |
| Index `WITH (...)` options | `pad_index`, `ignore_dup_key`, `statistics_norecompute`, `statistics_incremental`, `allow_row_locks`, `allow_page_locks` |

They belong to no keyword category, and the lexer's keyword table is shared across modes, so an unquoted column with one of these names fails to parse in every mode:

```
CREATE TABLE t (newid int)   -- syntax error in MySQL, PostgreSQL, SQLite3 and SQL Server modes
CREATE TABLE t (apply int)   -- same
```

A single such column makes the whole schema file unparsable, and in psqldef's default mode it silently falls back to pgquery, the same problem #1282 fixed for `TRIGGER` / `INDEX`. Quoted names (`[apply]`, `"apply"`) already work; only unquoted ones are affected.

As in #1282, this adds the 15 words to `non_reserved_keyword`. The dedicated rules such as `NEWID '(' ')'` stay, since `function_call_generic` only starts from `sql_id` (`ID`) and could not reach these tokens as function calls otherwise. The new parser test checks each word as a column definition, an index column and a column reference in all four modes. The mssqldef test creates a table and an index using all 15 words against a real SQL Server, which also confirms that SQL Server accepts them as unquoted names. The SQL Server syntax above is already covered by existing tests, except `GETUTCDATE()`, which is added to `TestDefaultFunctionExpressions`.

The conflict counts reported by `make parser` do not change (28 shift/reduce, 286 reduce/reduce).


moznion (approved):
LGTM

cmd/mssqldef/tests_preserve_quotes.yml
```diff
@@ -68,3 +68,48 @@ AddColumnBracketedIdentifier:
     ALTER TABLE dbo.users ADD [name] varchar(100);
   down: |
     ALTER TABLE dbo.users DROP COLUMN [name];
+
+CreateWithNonReservedKeywordColumns:
```
moznion, line 72: [quick question] do we need the test like this in other dialect?

178inaba: I don't think it's necessary.

The keyword table is shared by all four modes, and that is covered by `TestSQLServerKeywordsAsUnquotedIdentifiers`. The existing YAML tests also cover only the keywords of their own dialect (e.g. `CreateWithNonReservedKeywordColumns` in mysqldef covers only `definer`, `invoker` and `policy`).

That said, I don't feel strongly about it, so I'm happy to add the tests if you'd prefer.

moznion: OK, that sounds good enough. Let’s move forward with it.

178inaba: Thanks! Merged!

PR #1375, opened 2026-09-18
psqldef: add manage.owner for object ownership

gfx:
**On hold for redesign.** Ownership and view-recreation bug fixes have been extracted into [#1376](https://github.com/sqldef/sqldef/pull/1376), using the existing `manage.privilege` / `managed_roles` settings. This draft retains the `manage.owner` proposal for redesign, including reconsideration of the proposed v4 migration.

Adds `manage.owner` to control PostgreSQL ownership independently of privilege management. Rules match owner role names; `owner: []` manages all roles, while an omitted owner declaration leaves existing ownership unchanged. When `manage.owner` is omitted, both `manage.privilege` and `managed_roles` retain their existing ownership behavior; explicit owner rules take precedence.

Supports tables, partition parents and children, views, and materialized views, including `ALTER VIEW`, `ALTER MATERIALIZED VIEW`, and `IF EXISTS` owner statements. Public exports include only matching owners; internal diffs retain existing owners so view recreation can preserve ownership even for roles outside the configured scope.

View recreation now follows the current dependency graph for deletion and the desired graph for creation, restores ownership, and regenerates declared comments, privileges, and indexes. Includes fixes for skipped objects, identifier quoting, and held-back recreation with `enable_drop: false`, plus documentation and regression tests. The existing `ManagePrivilegeOwner*` and `ManagePrivilegeViewOwner` tests are retained with their original settings and expectations.

Validation:
- `make build`, `make format`, `make lint`, and `git diff --check` passed.
- PostgreSQL (generic parser), pgvector, MySQL, MariaDB, TiDB, SQLite, and core package tests passed; affected regression tests were rerun after the final fixes.
- `make test-all-flavors` could not complete because the local SQL Server container fails during startup under x86 emulation on this Mac. The remaining flavor suites were run separately.
- Parser regeneration reports 28 shift/reduce and 286 reduce/reduce conflicts, unchanged from both the branch's previous HEAD and the base parser.


gfx:
https://github.com/sqldef/sqldef/pull/1376 for now

PR #1377, opened 2026-09-20
mysqldef: stop bulk_alter from skipping non-destructive actions

178inaba:
## Why

`--bulk-alter` (#1201) was added to reduce DDL lock acquisitions to one per migration per table. It was never meant to change *which* changes get applied — but combined with `enable_drop: false`, the CLI default, it does.

`bulkAlter.finalize` fuses a table's `ALTER TABLE` actions before `commentOutDropStatements` gates them. A fused statement holding a single `DROP COLUMN` / `DROP INDEX` / `DROP PARTITION` therefore matches `isDropStatement` as a whole and is commented out entirely, stopping the `ADD COLUMN` actions bundled with it and the `DROP FOREIGN KEY` clauses the gate deliberately allows.

Two defects follow, both only with `--bulk-alter`:

1. **`--apply` never converges and still exits 0.** The column is never added, so every run prints the same line.

   ```
   $ mysqldef ... --apply --bulk-alter   # 1st and 2nd runs are identical
   -- Apply --
   -- Skipped: ALTER TABLE `patients` ADD COLUMN `a` int AFTER `id`, DROP COLUMN `c`;
   exit=0
   ```

2. **A schema adding a column plus a foreign key on it fails.** `ADD CONSTRAINT ... FOREIGN KEY` is a separate statement that is not fused, so it runs without its column. The same schema succeeds without `--bulk-alter`.

   ```
   $ mysqldef ... --apply --bulk-alter
   -- Apply --
   -- Skipped: ALTER TABLE `pets` ADD COLUMN `owner_id` bigint AFTER `id`, DROP COLUMN `c`;
   BEGIN;
   ALTER TABLE `pets` ADD CONSTRAINT `pets_owner_fk` FOREIGN KEY (`owner_id`) REFERENCES `owners` (`id`);
   ROLLBACK;
   Error 1072 (42000): Key column 'owner_id' doesn't exist in table
   exit=1
   ```

`bulk_alter` under `enable_drop: false` had no coverage, which is why this went unnoticed: YAML cases default to `enable_drop: true`, the opposite of the CLI default.

## What

- `alterBundler.emit` declines to bundle a statement `enable_drop` would gate, and returns it unchanged so that `commentOutDropStatements` — the single existing gate — comments it out. The remaining actions still fuse into one statement.
- The gate tests the whole statement rather than the action cut off the prefix. The action starts with `DROP ` for every destructive clause, so testing it would also catch the ones `isDropStatement` deliberately allows (`DROP FOREIGN KEY`, `DROP CONSTRAINT`, `DROP PRIMARY KEY`), changing what `enable_drop` means.
- The gate sits after the `strings.CutPrefix` check, so a statement that is not a plain `ALTER TABLE` for the table keeps passing straight through to the final gate.

With this, `--bulk-alter` produces the same output as a run without it, down to the `-- Skipped:` lines:

```
$ mysqldef ... --apply --bulk-alter
-- Apply --
BEGIN;
ALTER TABLE `pets` ADD COLUMN `owner_id` bigint AFTER `id`;
ALTER TABLE `pets` ADD CONSTRAINT `pets_owner_fk` FOREIGN KEY (`owner_id`) REFERENCES `owners` (`id`);
-- Skipped: ALTER TABLE `pets` DROP COLUMN `c`;
COMMIT;
exit=0
```

Safe actions still fuse when there is more than one:

```
ALTER TABLE `patients` ADD COLUMN `a` int AFTER `id`, ADD KEY `a_idx` (`a`);
-- Skipped: ALTER TABLE `patients` DROP COLUMN `c`;
```

The `enable_drop: true` output is unchanged, and a statement skipped on this path keeps its `ALGORITHM` / `LOCK` suffix, since it now reaches those passes exactly as it does without `--bulk-alter`.

## Alternatives considered

- **Swapping `finalize` and `commentOutDropStatements`.** Does not work: `emit` leaves only an `ALTER TABLE t ` prefix placeholder at the first action's position, so the actions are invisible to a gate that runs before `finalize`.
- **Commenting the statement out inside `emit`.** Works, but puts the rendering of `-- Skipped: ` in a second place, and the early comment no longer matches the `ALTER TABLE` prefix the `ALGORITHM` / `LOCK` passes look for — so `bulk_alter` would change the text of statements it explicitly refuses to bundle.
- **Rejecting the whole `ALTER` with an error when any action is destructive.** `enable_drop` has meant "skip the destructive statements, apply the rest" since #682, and this would also disagree with the behaviour without `--bulk-alter`. A contract change does not belong in a bug fix.
- **Carrying destructiveness as a typed value** instead of inferring it from the generated SQL text. That is the shape this would take if written from scratch, but it spans the whole generator and is left out deliberately.

## Not covered

- **A statement commented out at emission time carries no `ALGORITHM` / `LOCK` suffix.** The suffix passes match on the `ALTER TABLE` prefix, which a `-- Skipped: ` line no longer has. The path this PR touches reaches those passes normally, but the pre-existing emission-time gates — `appendRecreate`, `manage.privilege`, `manage.function` — do not, so their skipped lines still print without the suffix. Comment text only; nothing about what runs changes.
- **`DROP INDEX` is not bundled at all in quote-aware mode when the input DDL is unquoted.** The cleanup `DROP INDEX` carries the current table's name, which always comes quoted from the database, so it does not match the bundle's prefix built from the desired name. This is independent of `enable_drop` and does not change which changes get applied — only that one statement is not fused — so it is left for a separate change. `BulkAlterDropIndexSkipped` therefore uses quoted identifiers; with unquoted ones no fusing happens and the case would not reproduce the bug.

## Tests

Four YAML cases in `cmd/mysqldef/tests_bulk_alter.yml`, all `enable_drop: false` with `bulk_alter: true`. All of them fail before this change:

- `BulkAlterDropColumnSkipped`: adding a column plus dropping one.
- `BulkAlterDropIndexSkipped`: adding a column plus dropping an index.
- `BulkAlterDropForeignKeyDropColumnSkipped`: `DROP FOREIGN KEY` is applied while the `DROP COLUMN` in the same table is skipped.
- `BulkAlterAddIndexDropColumnSkipped`: two safe actions fuse while the `DROP COLUMN` is skipped. Marked `flavor: "!tidb"`, because the fused `ADD KEY` references the column added in the same statement.

`TestAlterBundlerSkipsDropsWhenDropDisabled` in `schema/generator_test.go` pins the bundler itself: a gated `DROP COLUMN` is not bundled, a `DROP FOREIGN KEY` still folds, and the safe actions fuse into one statement. `TestAlterBundler` now runs with `EnableDrop: true`.

`TestMysqldefConfigIncludesAlgorithmAndLockOnBulkAlterSkipped` in `cmd/mysqldef/mysqldef_test.go` pins the `ALGORITHM` / `LOCK` suffix on a skipped drop under `bulk_alter`. The YAML config has no `algorithm` / `lock` keys, so this one has to be a Go test.

`make test-all-flavors`, `make format` and `make lint` pass.


moznion (approved):
LGTM, thank you!

PR #1378, opened 2026-09-21
psqldef: match unnamed indexes and constraints by definition

178inaba:
## Why

psqldef names an unnamed UNIQUE constraint, index, PRIMARY KEY, or foreign key by imitating PostgreSQL's naming rules, and uses that name both to match the current schema and in the DDL it emits. Wherever the guess differs from the name PostgreSQL actually chose, psqldef mistakes one object for another. Applying each schema below to an empty database on master (f1e3a54, PostgreSQL 18):

- **A name taken by another index sends psqldef into a loop.** With `CREATE INDEX t_a_idx ON t (b); CREATE INDEX ON t (a);`, PostgreSQL would name the second index `t_a_idx1`, but psqldef takes it to be the existing `t_a_idx`, which has a different definition. Without `--enable-drop`, the index on `a` is never created (`-- Skipped: DROP INDEX "public"."t_a_idx"` on every run). With `--enable-drop`, every run drops the index on `b`, recreates it, drops it again, and creates the index on `a`, so each run ends without the index on `b` and the schema never converges. With this PR, both indexes are created (`t_a_idx` on `b`, `t_a_idx1` on `a`) and the second run is empty.
- **Multibyte names break the schema permanently.** A `UNIQUE` on a table and column named in Japanese is created fine, but the second run cuts the predicted name in the middle of a character and fails with `invalid byte sequence for encoding "UTF8"`. Every later run fails the same way.
- **Long names are re-created.** For a `UNIQUE` on a 40-byte table name and a 40-byte column name, psqldef truncates the name at a different position than PostgreSQL, so the second run emits `ADD CONSTRAINT "tttt…_cccc…_key"` and `DROP CONSTRAINT "tttt…_cccc…_key"` with different truncations, and both run (`DROP CONSTRAINT` is not gated by `--enable-drop`).
- **Unquoted mixed-case identifiers** (`legacy_ignore_quotes: false`): `UNIQUE (Stateful)` is re-created the same way on the second run, and `CREATE INDEX ON t (Stateful)` creates an index named `"t_Stateful_idx"` that the schema file never mentions.
- **Renaming a table rebuilds its unnamed indexes.** After applying `CREATE TABLE t (a int); CREATE INDEX ON t (a);`, renaming the table with `CREATE TABLE u ( -- @renamed from=t` and `CREATE INDEX ON u (a);` emits `ALTER TABLE "public"."t" RENAME TO "u"; CREATE INDEX ON u (a); DROP INDEX "public"."t_a_idx";` under `--enable-drop`. PostgreSQL keeps the name `t_a_idx` across the rename, while psqldef now expects `u_a_idx`. With this PR, only the `RENAME` is emitted and the index stays as `t_a_idx`.

Fixing the prediction one divergence at a time would not converge. The counter PostgreSQL appends on a collision depends on what is already in the database, so it cannot be predicted from the schema file. PostgreSQL also keeps a name once it has chosen it, so even an exact emulation run against the current catalog gives the wrong name after a table is renamed or the colliding object is gone. The rest follows implementation details of PostgreSQL (`ChooseIndexName`, `makeObjectName`) that can change between versions.

## What

Following what #1324 did for CHECK constraints, an unnamed declaration is matched to the current schema by its definition, and one that has to be added is emitted without a name so that PostgreSQL chooses it. The code that predicted PostgreSQL's names is removed. Other dialects are unchanged.

- Each table and materialized view gets a match plan, like `postgresCheckMatchPlan`: named desired indexes are paired with current ones by name first, then unnamed ones by `areSameIndexes`. Each current index is paired at most once, and current indexes are scanned in name order so the output is deterministic. CHECK and indexes now share this pairing (`matchByNameThenDefinition`).
- Every path that handles indexes uses the same plan: the index loop of `CREATE TABLE`, each `CREATE INDEX` / `ALTER TABLE ADD` statement, and the cleanup of absent indexes. The plan is taken from the current indexes as they were before this run added any, so an index added in this run is never mistaken for an existing one.
- `ALTER TABLE ... ADD FOREIGN KEY` without a name is matched by columns, as foreign keys in `CREATE TABLE` already were.
- Obsolete indexes on materialized views are now dropped, like those on tables. Without this, changing the definition of an unnamed index on a materialized view would leave the old index behind even with `--enable-drop`, since the old and new definitions no longer share a predicted name.

### Behavior changes

- DDL that adds an unnamed declaration no longer carries a name: `CREATE INDEX ON ...`, `ADD UNIQUE (...)`, `ADD PRIMARY KEY (...)`, `ADD FOREIGN KEY ...`. The name PostgreSQL gives it is its default, which is the same name as before wherever the prediction was right.
- An unnamed declaration matches a current object with the same definition regardless of that object's name. An index the user named is not re-created when an unnamed declaration has the same definition, and removing only the name from the schema file emits nothing. This is the same as CHECK in `ReplaceUnnamedCheckWithExplicitName` (`down: ""`).
- Unnamed declarations with the same definition are counted: two `CREATE INDEX ON t (a)` stand for two indexes. Of current indexes with the same definition, the one first by name is kept.
- Changing the definition of an unnamed index treats the old one as an index absent from the desired schema. Without `--enable-drop`, the new index is created and the `DROP INDEX` of the old one is skipped; both used to be held back. This is what already happens when a named index is given a different name.
- A named index removed from a materialized view's schema is now dropped; it used to be left in place.
- When the current schema is read from a file, dropping an unnamed index or UNIQUE constraint is an error, as it is for CHECK.
- A database that still has objects under previously predicted names (`"t_Stateful_idx"` etc.) emits no DDL, since they match by definition.
- An unnamed foreign key whose reference changed is added as a new foreign key and the old one is dropped; it used to be recreated under the predicted name. `ALTER TABLE ... ADD FOREIGN KEY` without a name used to be re-added on every run instead.
- Unnamed index constraints repeated with the same definition in one `CREATE TABLE` count as one, since PostgreSQL creates a single index for them.

## Alternatives not taken

- **Keep predicting and fix each divergence.** The collision counter cannot be predicted, and a database that already has objects under the old names would keep two indexes with the same definition until one run with `--enable-drop`, failing `--check` meanwhile. Matching by definition does not have this problem.
- **Match by definition but still emit the predicted name.** The spurious re-creation goes away, but indexes would still be created under names the schema file never mentions (`"t_Stateful_idx"`), and the prediction code would stay.

## Not changed

- Named declarations are still matched by name only: a named desired index whose definition exists only under another name is added and the other one dropped, unless `@renamed` says otherwise.
- `@renamed` on an unnamed index: there is no name for it to refer to. The documentation now says to name an index that is referred to by name.
- The PRIMARY KEY comparison (`areSamePrimaryKeys` does not compare names) and the DDL that adds one.

## Tests

The new YAML cases cover the cases above, each applied twice to check that the second run is empty. `TestPostgresIndexMatching` fixes the pairing rules without a database, including two things a YAML case cannot check: current indexes scanned in name order (the export already returns same-definition indexes in name order), and an unnamed current index read from a file.

Tests whose expectations changed are all `up`/`down` DDL that adds an unnamed declaration, which is now emitted without a name. DROP expectations are unchanged.

- `tests_constraints.yml`: `ConstraintUniqueAdd`, `ConstraintUniqueRemove`, `ConstraintCheckInAndUniqueAdd`, `ConstraintCheckInAndUniqueRemove`, `ConstraintCheckInMultipleColumnsWithUnique`, `CreateTableAddAbsentForeignKey`, `DropReferences`, `ForeignKeyConstraintsAreEmittedLast`, `ForeignKeyOnReservedName`
- `tests_preserve_quotes.yml`: `ForeignKeyAutoNameWithQuotedColumn`
- `tests_indices.yml`: `UnnamedIndexWithAccessMethod`

`TestAutoIndexName` loses its PostgreSQL cases.

<details>
<summary>Places keyed by index name</summary>

With PostgreSQL, an unnamed declaration's `Index.name` is now empty. I went through the places that key on an index or foreign key name.

Fixed:

- `findForeignKeyByName`: with PostgreSQL, an empty name never matches (SQLite keeps unnamed foreign keys and still matches them this way).
- `mergeTable`: an unnamed index is appended rather than deduplicated by name; one that matched a current index carries that index's name.
- `sortIndexesByName`: stable, so indexes with the same name keep their order.
- `generateAddIndex`, `generateCreateIndexStatement`, `generateForeignKeyDefinition`: omit the name when it is empty.
- `isAddConstraintForeignKey`: also recognizes `ADD FOREIGN KEY`, so unnamed foreign keys are still emitted last.
- The foreign key recreation after a primary key change looks the desired foreign key up with `findForeignKeyByName`.
- `findMatchingDesiredForeignKey`: with PostgreSQL, a current foreign key is kept only if an unnamed desired one has the same reference, not just the same source columns.

Read and kept:

- `findIndexByName`: never receives an empty name, since unnamed indexes are matched through the plan.
- `replaceIndex`, `trackHeldBackIndex`, `generateRenameIndex`: only reached for named indexes; an unnamed one that matched carries the current name.
- `trackDroppedIndex`, `droppedIndexKey`: only reached with a name, since dropping an unnamed index is an error.
- `buildIndexToTableMap`: an index from the database always has a name; an unnamed one from a file is never the target of `COMMENT ON INDEX`.
- `indexComments[nameIdent.Name]` in `schema/parser.go`: `extractIndexComments` never stores an empty key.
- `convertForeignKeysToIndexNames`: only returns non-empty names.
- `handledForeignKeys`: never set under an empty name, since the lookup above does not match one. A primary key change with an unnamed referencing foreign key was checked to re-add it when the referencing table is processed.

</details>


moznion (approved):
Thank you for your contribution. Overall, it looks good but I suppose it would be nicer if the following regression tests are covered in `tests_indices.yml`:

```yaml
UnnamedIndexWhoseDefaultNameIsTaken:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE t (a int, b int);
  desired: |
    CREATE TABLE t (a int, b int);
    CREATE INDEX t_a_idx ON t (b);
    CREATE INDEX ON t (a);
  up: |
    CREATE INDEX t_a_idx ON public.t (b);
    CREATE INDEX ON public.t (a);
  down: |
    DROP INDEX public.t_a_idx;
    DROP INDEX public.t_a_idx1;

RenameTableKeepsUnnamedIndex:
  legacy_ignore_quotes: false
  current: |
    CREATE TABLE t (a int);
    CREATE INDEX ON t (a);
  desired: |
    CREATE TABLE u ( -- @renamed from=t
      a int
    );
    CREATE INDEX ON u (a);
  up: |
    ALTER TABLE public.t RENAME TO u;
  down: |
    CREATE TABLE t (a int);
    CREATE INDEX ON public.t (a);
    DROP TABLE public.u;
```

